PR #363, opened 2018-11-25
Pin to specific rust nightly version

badboy:
This is automatically used by rustup.


pietroalbini:
Fixed in https://github.com/rust-lang/www.rust-lang.org/pull/712. Closing this.

PR #395, opened 2018-11-29
Describe rustup as "version management" instead of "version control"

porglezomp:
Version control generally means something different, so I did a double take when I first read it. I think "version management" expresses what it's for clearly without having another established meaning.


ashleygwilliams:
thank you!

PR #433, opened 2018-11-29
s/rock-solid compiler/ownership model/

oconnor663:
Closes https://github.com/rust-lang/beta.rust-lang.org/issues/409.

It looks like https://github.com/rust-lang/beta.rust-lang.org/pull/403
already addressed the first suggestion in that issue, so this commit
only touches on the second.

cc @aturon


aturon:
Perfect, thank you!

PR #453, opened 2018-11-29
addition of "extern crate ferris_says" to the example

shef:
I was not able to build the program without `extern crate ferris_says`, the build was failing with error[E0432]: unresolved import `ferris_says`


ashleygwilliams:
we don't add the `extern crate` because that is not needed in the 2018 edition. for previous editions you will need it- but as of dec 6th, when the site goes live on rust-lang.org, that will be the default.

PR #465, opened 2018-11-30
fix #429

### LINKED ISSUE #429: Handle a JSON POST - Missing Angle Brackets
# Content Bug Report

## Page(s) Affected
Networking -> Handle a JSON POST

## What needs to be fixed?
The code snippet's angle bracket sections are missing (see the method signature):
```rust
#[derive(FromForm)]
struct Task { name: String, completed: bool }

#[post("/", data = "")]
fn new(task: Form) -> Flash {
    if task.name.is_empty() {
        Flash::error(Redirect::to("/"),
            "Cannot be empty.")
    } else {
        Flash::success(Redirect::to("/"),
            "Task added.")
    }
}
```

This appears to be taken from the Rocket front page example, which is correct.

## Suggested Improvement
The method signature should be:
```rust
#[post("/", data = "<task>")]
fn new(task: Form<Task>) -> Flash<Redirect>
```

nikhiljha:
There were angle brackets in the original handlebars file, but they were being parsed out (the parser thinks its HTML). I replaced them with escaped HTML versions so that they show up correctly.

### DIFF
--- a/templates/what/networking/take-json.hbs
+++ b/templates/what/networking/take-json.hbs
@@ -1,8 +1,8 @@
 #[derive(FromForm)]
 struct Task { name: String, completed: bool }
 
-#[post("/", data = "<task>")]
-fn new(task: Form<Task>) -> Flash<Redirect> {
+#[post("/", data = "&lt;task>")]
+fn new(task: Form&lt;Task>) -> Flash&lt;Redirect> {
     if task.name.is_empty() {
         Flash::error(Redirect::to("/"),
             "Cannot be empty.")


nikhiljha:
Hmm, [that doesn't seem to work](https://stackoverflow.com/questions/34339493/handlebars-include-partial-twice-once-escaped-and-once-rendered). I agree that the current PR is a hacky way to do it though.

ashleygwilliams:
hrm i'm not sure i follow that stack overflow issue... did you play with it? i may try it out later today and just see- as that should be the correct way to do it.

ashleygwilliams:
now that we've merge the update to rocket- could you try again with the `{{{`?

yoshuawuyts:
ping @nikhiljha - are you still interested in landing this patch?

nikhiljha:
There's no way to do an escaped partial in handlebars without writing custom logic, so I guess not.

The simplest solution that I can find (at the expense of some readability) is the one in this PR.

yoshuawuyts:
@nikhiljha ah, that's unfortunate. We're very grateful for the time you've invested exploring this!

PR #469, opened 2018-11-30
fixed typo in Ferris section of get started page

Maxgy:
I fixed the "which is set of creatures" phrase in the Ferris section at the bottom of the Get Started page.  It had a typo where it was missing an 'a'.


aturon:
Thanks!

PR #473, opened 2018-11-30
Fix the example of small Rust application by adding 'extern crate ferris_says;' as a first line

kodersky:


wezm:
`extern crate` isn't needed in the 2018 edition, which the website is intended to convey.

https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/path-clarity.html#no-more-extern-crate

kodersky:
Ok. Thanks.

PR #494, opened 2018-11-30
feat(community): add irc link

ashleygwilliams:
fixes #492 

small github dance because the previous PR was merged too soon

screen shot 2018-11-30 at 11 51 26 am


alyssais:
This still makes IRC look like a second-class citizen, I think? It’s below Discord, and the text outright says to join the Discord, despite that IRC is supposedly still the official platform.

PR #497, opened 2018-11-30
feat(nav): add install link

### LINKED ISSUE #459: add install to top nav
# Design Feature Request
## Summary
AFAICT, the "Getting Started" page is only on the home page. I propose we add it as a link in the navbar at the top of every page.

## Motivation
If new potential users cannot find this link, they may just give up.

## Drawbacks
It would be an additional link at the top of the page. I think this is worth it.

## Rationale and alternatives
This proposal is a simple change, which I think should have minimal disturbance of the current design.

Alternatives
- Add a link at the bottom of every page, say in the footer or as a last panel at the bottom of each page. Putting it in the footer might still be hard to find. Putting it as a panel at the bottom might take more design effort.
- Put it at the top of each page but not in the navbar. This might disrupt the current visual design and require some work.

## Maintenance
None necessary

## Unresolved Questions
None TMK

ashleygwilliams:
fixes #459 

screen shot 2018-11-30 at 12 45 07 pm
screen shot 2018-11-30 at 12 46 23 pm

not great on mobile but not terrible, cc @chriskrycho

### DIFF
--- a/templates/nav.hbs
+++ b/templates/nav.hbs
@@ -8,6 +8,7 @@
   </a>
 
   <ul class="nav list w-100 w-auto-l flex flex-none flex-row flex-wrap justify-center justify-end-l items-center pv2 ph4-ns">
+    <li class="flex-inline pv2 ph2 ph4-ns"><a href="/tools/install">Install</a></li>
     <li class="flex-inline pv2 ph2 ph4-ns"><a href="/learn">Learn</a></li>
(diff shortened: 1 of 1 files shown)


aturon:
I'm :+1: on this change! 

We may want to also consider having the /tools/install page have a button to go to the longer "getting started" page to help walk people through.

PR #501, opened 2018-11-30
Add blurb before domains

### LINKED ISSUE #458: Highlighting 4 working group areas may discourage other use cases
# Content Feature Request
## Summary
Add at least a paragraph explaining that the "Build it with Rust" section of the home page is just a sampling of things that we know work well. In particular, there are great crates for everything from SIMD to graphics. A link to crates.io's categories page might be helpful too.

## Motivation
Highlighting only the four areas with working groups is a great way to help common use cases. However, the fact that no other use cases are mention IMHO implies that you can pretty much only do those 4 things. This is not a message we should convey, I think.

## Drawbacks
More text/clutter on the website, though I think it's worth it.

## Rationale and alternatives
The proposed change is a minimal change to the existing design that should be possible with minimal effort. It's of course possible to not do anything, but that has the stated downsides.

## Maintenance
None necessary.

## Unresolved Questions
None, TMK

mark-i-m:
I also added a blurb before the WG spotlights. The crates.io image comes from the rust-lang/crates.io repo (public/cargo.png). Also, I'm not familiar with the stylesheets being used, so please let me know if I am using the wrong classes or something.

EDIT: as per https://github.com/rust-lang/beta.rust-lang.org/pull/501#pullrequestreview-180441720 , I am just adding a the blurb for now...

r? @aturon


mark-i-m:
Oops... forgot to include a screenshot

screenshot_2018-11-30 rust - index

templates/panels/domains.hbs
```diff
@@ -7,6 +7,38 @@
...
+            and "Popular Categories" lists to see the most popular crates or
+            search around and explore for yourself!
+          </p>
+        </div>
+      </div>
+    </div>
+
+    <div class="flex-none flex-l flex-row">
+      <p class="flex-grow-1">
+        In 2018, the Rust community decided to improve programming experience
```
ashleygwilliams: we might want to link to the 2018 roadmap here!

mark-i-m:
> but i dont think the crates.io callout works. for the sake of small PRs perhaps we can make this PR just the sentence, and figure out crates.io in another issue?

sure, perhaps we can continue discussion on #458?

I removed all but the sentence.

ashleygwilliams:
@aturon can you review the sentence? i think we may want some tweaks but i'm going to leave that in your court :)

PR #513, opened 2018-12-01
Fix clippy warnings, fix #511

### LINKED ISSUE #511: Run clippy on source
# Technical Bug Report

## Is this an issue with Rust, CSS, or HTML (or a combination)?
Issue with Rust source.

## Expected Behavior
`cargo clippy` should not output anything.
Unless the team decides not to use/take into account clippy.

## Current Behavior
We have a few cargo clippy warnings that can help have idiomatic rust code.
And with all clippy outputs you get a simple way to fix the warning.

## Step to Reproduce
run `cargo clippy`

meven:
Fix #511 

This simplify the code a little and avoid formatting string in expect blocks.


badboy:
Probably because that defers actually calling the format function (or any function for that matter).

PR #518, opened 2018-12-01
Mention `rustup upgrade` in install instructions

sophiebits:
I had installed `rustup` a while ago but sat down today to try Rust again; it wasn't obvious to me how to get the latest Rust.

### DIFF
--- a/templates/tools/install.hbs
+++ b/templates/tools/install.hbs
@@ -40,6 +40,10 @@
         cross-compilation targets.
       </p>
       <p>
+        If you've installed <code>rustup</code> in the past, you can update
+        your installation by running <code>rustup upgrade</code>.
+      </p>
+      <p>
         For more information see the
         <a href="https://github.com/rust-lang-nursery/rustup.rs/blob/master/README.md" target="_blank" rel="noopener"><code>rustup</code>
         documentation</a>.


templates/tools/install.hbs
```diff
@@ -40,7 +40,9 @@
         cross-compilation targets.
       </p>
       <p>
-        For more information see the
+        If you've installed <code>rustup</code> in the past, you can update
```
chriskrycho: One suggestion: let's put this in its own paragraph, as the other blocks are.

```hbs
<p>
  If you've installed <code>rustup</code> in the past, you can update
  your installation by running <code>rustup upgrade</code>.
</p>
```

Then we can leave the original in place.

sophiebits: Sure, if you prefer. Updated.

chriskrycho: Thank you!

meven:
I don't know this "update" subcommand shouldn't is be "update"

>  rustup -h
> rustup 1.15.0 (f0a3d9a84 2018-11-27)
> The Rust toolchain installer
> 
> USAGE:
>     rustup [FLAGS] <SUBCOMMAND>
> 
> FLAGS:
>     -v, --verbose    Enable verbose output
>     -h, --help       Prints help information
>     -V, --version    Prints version information
> 
> SUBCOMMANDS:
>     show           Show the active and installed toolchains
>     update         Update Rust toolchains and rustup
>     default        Set the default toolchain
>     toolchain      Modify or query the installed toolchains
>     target         Modify a toolchain's supported targets
>     component      Modify a toolchain's installed components
>     override       Modify directory toolchain overrides
>     run            Run a command with an environment configured for a given toolchain
>     which          Display which binary will be run for a given command
>     doc            Open the documentation for the current toolchain
>     man            View the man page for a given command
>     self           Modify the rustup installation
>     set            Alter rustup settings
>     completions    Generate completion scripts for your shell
>     help           Prints this message or the help of the given subcommand(s)
>

chriskrycho:
Oh, great catch! Do you have a few minutes to open a PR fixing that word, @meven? I'll happily review and merge it!

meven:
PR opened ;)

sophiebits:
That’s embarrassing! The choice to use “update” in prose was intentional but I meant to write “upgrade” in the command. Thanks for fixing!

Issue #446, opened 2018-11-29
Reduce padding on the hompage under the slogan


nicoburns:
# Reduce padding on the hompage
## Summary

Reduce padding / whitespace on the homepage

## Motivation

The current homepage doesn't not make effective use of space:

<img src="https://user-images.githubusercontent.com/3943692/49254345-ef8e2080-f420-11e8-8e77-5d918b90447a.png" />

ashleygwilliams:
i would happily accept a PR that reduces the white space under the slogan! thanks!

nicoburns:
If the PR above is accepted, then the homepage will look like this:

<img width="1440" alt="screenshot 2018-11-29 at 22 49 05" src="https://user-images.githubusercontent.com/1007307/49257080-0a649300-f429-11e8-8191-754664fa0ff3.png">

ssokolow:
Having so little padding relative to the font size feels slightly cramped to me

Perhaps a *little* more padding below the slogan... more like the padding below the word "more" in the "Productivity" block?

nicoburns:
@ssokolow I got the same feedback from Ashley. Final design that got merged looks like this:

![localhost_7878_ copy](https://user-images.githubusercontent.com/1007307/49259922-2a994f80-f433-11e8-95d2-58322f110a64.png)

ssokolow:
Looks good. :)

(Well, I have other issues with the design, but the padding is no longer one of them.)

ashleygwilliams:
I'm reopening this because I needed to revert the previous PR because it caused issues on pages with subtitles or long titles. i am very interested in a PR that would improve this- just we need to make sure it works for all pages. Sorry for the back and forth- i'm happy to work on this with someone if they'd like to take this on!

ashleygwilliams:
@chriskrycho if you end up with some extra time, this would be a good one to check out :)

chriskrycho:
Top of my list for Monday!

nicoburns:
@chriskrycho I think this fixes the immediate issue with the previous PR, by replacing the tachyons class with padding defined in app.scss.

I'm unsure of what the overall strategy for CSS is, and I've never worked with tachyons or a similar framework before. But won't it always have this problem? That changes need to be applied individually to every page (or moved out of the tachyons system). Is there any plan to complement tachyons with an ITCSS framework like Inuit.css?

chriskrycho:
Hey @nicoburns – thanks for submitting another PR!

[...]

PR #526, opened 2018-12-02
Reduce homepage whitespace take 2

nicoburns:
Another stab at #446. Adds the header padding to all pages.


ashleygwilliams:
can you include some screencaps, including pages with subtitles (governance is an example) and long titles (command line is an example) as well as the landing page? thanks so much!

Issue #447, opened 2018-11-29
user logos should be vertically centered on networking page


jonas-schievink:
# Content Bug Report

## Page(s) Affected
http://beta.rust-lang.org/what/networking

## What needs to be fixed?

The Buoyant logo is placed in a way that makes it look like it belongs to the quote above (which is from Mozilla), not the correct quote on the left of it:

![screenshot-2018-11-29-23 41 12](https://user-images.githubusercontent.com/1786438/49256737-4818ea00-f430-11e8-8a5c-a4bbc129e892.png)

## Suggested Improvement
The logo should be vertically centered in its space and should be at the same height as the text it belongs to. Also consider reducing the distance between text and logo (it's rather large for the Firefox logo).

ashleygwilliams:
i would happily accept a PR to fix this! thank you

jonas-schievink:
Someone else will have to implement this, I've spent an hour wrangling CSS and couldn't get it to align properly without breaking responsiveness

ashleygwilliams:
closed by #530

PR #528, opened 2018-12-03
fix(production.hbs): Vertically center logos in testimonials

dustinknopoff:
fix for #447

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -623,3 +623,9 @@ blockquote::before {
   display: flex;
   align-items: center;
 }
+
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
could you include a screencap? thanks so much @dustinknopoff !

dustinknopoff:
screen shot 2018-12-03 at 12 46 17 pm

ashleygwilliams:
looks great! there are a few other places where vertically centering would be great! i'll merge this but would also welcome PRs for the users page (and other pages with logos)!

PR #540, opened 2018-12-04
feat(css): concat vendor css

### LINKED ISSUE #538: concat and min vendor stylesheets
we currently load 2 vendor stylesheets (which won't be forever) but while we are in this in between spot, let's concat the 2 vendor stylesheets so the load flickering isn't as bad.

ashleygwilliams:
fixes #538


src/main.rs
```diff
@@ -238,9 +239,23 @@ fn compile_sass(filename: &str) {
         .expect(&format!("couldn't write css file: {}", &css_file));
 }
 
+fn concat_vendor_css(files: Vec<&str>) {
+    let mut concatted = String::new();
+    for filestem in files {
+        let vendor_path = format!("./static/styles/{}.css", filestem);
+        let mut file = File::open(vendor_path).expect("couldn't read vendor css");
```
chriskrycho: Certainly not blocking on this, just a nice trick: for this exact pattern you can use [`fs::read_to_string`](https://doc.rust-lang.org/1.30.1/std/fs/fn.read_to_string.html) instead:

```rust
let contents = fs::read_to_string(vendor_path).expect("couldn't read vendor css");
concatted.push_str(&contents);
```

PR #585, opened 2018-12-06
fix #582

### LINKED ISSUE #582: cli "Get started!" code snippet does not compile
# Content Bug Report

## Page(s) Affected
Please list the page or pages this bug is present on. Please file one issue
for each bug you find. Only file an issue for a multi-page bug if the bug is
indeed the *same* bug, across the pages.

https://www.rust-lang.org/what/cli

## What needs to be fixed?
Tell us what content is causing the issue and *how* it is causing the issue.
For example, "The code sample was confusing because it used advanced Rust features",
or "The call to action on the community page was vague".

The type of "num" from struct Cli because the method "take" accepts "usize" as an argument.
"let args = Cli::from_args()?:" should be "let args = Cli::from_args()?;"
Should also add "use structopt::StructOpt;"

## Suggested Improvement
(optional) Do you have a suggestion for how to fix the problem? Include your ideas
here.

Change "num" in struct Cli from "i32" to "usize".
Change "let args = Cli::from_args()?:" to "let args = Cli::from_args()?;"     ":" -> ";"
Add "use structopt::StructOpt;"

andreilorin:
fix typo.
change field type.
add crate.

### DIFF
--- a/templates/what/cli/code-inputs.hbs
+++ b/templates/what/cli/code-inputs.hbs
@@ -5,5 +5,5 @@ struct Cli {
     file: String,
     /// Number of lines to read
     #[structopt(short = "n")]
-    num: i32,
+    num: usize,
 }
--- a/templates/what/cli/code-main.hbs
+++ b/templates/what/cli/code-main.hbs
@@ -1,7 +1,8 @@
 use quicli::prelude::*;
+use structopt::StructOpt;
 
 fn main() -> CliResult {
-    let args = Cli::from_args()?:
+    let args = Cli::from_args()?;
     read_file(&args.file)?
         .lines()
         .take(args.count)


andreilorin:
In order for the example code to compile anyone can just copy both snippets into the same file and run them. 
rustc 1.31.0

andreilorin:
This was my first open source pull request... It made me really happy that it got approved !
Thank you!

killercup:
Thanks so much for this! I *just* realized that I forgot to add the structopt import to the code snippet on the website and was about to make a PR -- only to see you already fixed it! <3

andreilorin:
I'm glad I was able to make a small contribution and learn a bit of rust along the way :-)

PR #596, opened 2018-12-07
reduce highlight block height from 30px to 24px

stevensonmt:
highlightboxchangeproposal
height change necessitates also changing `top` positioning to -24px in order to maintain alignment relative to bottom of text.

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -289,9 +289,9 @@ header h1 {
 }
 
 .highlight {
-  height: 30px;
+  height: 24px;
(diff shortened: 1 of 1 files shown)


stevensonmt:
The height of the current highlight block cuts through the whitespace of the letter forms, making it much more difficult to read. Reducing the height of the highlight block to avoid interrupting that whitespace improves the readability of the section titles. At the same time, this change preserves the bold, blocky, contrast design.

nrc:
Thanks! LGTM!

PR #597, opened 2018-12-07
Add redirects to keep old URIs working

### LINKED ISSUE #593: URLs for the old website should redirect (if appropriate)
For example, https://www.rust-lang.org/conduct.html to https://www.rust-lang.org/policies/code-of-conduct instead of 404. Let’s try not to break existing links :)

mbrubeck:
This will allow existing links to the Rust code of conduct and other pages to keep working, instead of returning 404 errors.  Fixes #593 .

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,6 +14,7 @@ extern crate serde_derive;
 mod category;
 mod group;
 mod production;
+mod redirect;
 mod rust_version;
 
 use group::*;
@@ -28,7 +29,7 @@ use std::path::{Path, PathBuf};
 
 use rand::seq::SliceRandom;
 
-use rocket::response::NamedFile;
+use rocket::response::{NamedFile, Redirect};
 use rocket::Error;
 use rocket_contrib::Template;
 
@@ -222,6 +223,22 @@ fn subject(category: Category, subject: String) -> Template {
     Template::render(page, &context)
 }
 
+#[get("/<dest>", rank = 9)]
+fn redirect(dest: redirect::Destination) -> Redirect {
+    Redirect::permanent(dest.uri)
+}
+
+#[get("/en-US/<dest>")]
+fn redirect_en_us(dest: redirect::Destination) -> Redirect {
+    Redirect::permanent(dest.uri)
+}
+
+#[get("/<_locale>/<dest>", rank = 9)]
+fn redirect_locale(_locale: redirect::Locale, dest: redirect::Destination) -> Redirect {
+    // Temporary until locale support is restored.
+    Redirect::temporary(dest.uri)
+}
+
 #[catch(404)]
 fn not_found() -> Template {
     let page = "404";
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -222,6 +223,16 @@ fn subject(category: Category, subject: String) -> Template {
...
 }
 
+#[get("/<dest>", rank = 9)]
+fn redirect(dest: redirect::Destination) -> Redirect {
+    Redirect::permanent(dest.uri)
+}
+
+#[get("/<_locale>/<dest>", rank = 9)]
+fn redirect_locale(_locale: redirect::Locale, dest: redirect::Destination) -> Redirect {
+    Redirect::permanent(dest.uri)
```
pietroalbini: I'd prefer to avoid permanent redirects, especially for the locale links. For example, if we add translations back into the website in the future it would be nice to redirect old links to the correct locale.

mbrubeck: Changed to use temporary redirects for links to non-en-US locales.

Soria0906:
Ok

Carreau:
Do you want to log the referrer if possible when redirect, so that some could get fixed (like stackoverflow)?

src/redirect.rs
```diff
@@ -0,0 +1,74 @@
...
+            "es-ES" => Ok(Locale("es-ES")),
+            "fr-FR" => Ok(Locale("fr-FR")),
+            "id-ID" => Ok(Locale("id-ID")),
+            "it-IT" => Ok(Locale("it-IT")),
+            "ja-JP" => Ok(Locale("ja-JP")),
+            "ko-KR" => Ok(Locale("ko-KR")),
+            "pl-PL" => Ok(Locale("pl-PL")),
+            "pt-BR" => Ok(Locale("pt-BR")),
+            "ru-RU" => Ok(Locale("ru-RU")),
+            "sv-SE" => Ok(Locale("sv-SV")),
```
VladimirMarkelov: I am not sure but it looks suspicious: locale `sv-SV` for param `sv-SE`. It is the only line with different param/locale pair.

mbrubeck: Fixed, thanks.

src/redirect.rs
```diff
@@ -0,0 +1,74 @@
...
+            "downloads.html" => "/tools/install",
+            "friends.html" => "/production",
+            "index.html" => "/",
+            "install.html" => "/tools/install",
+            "legal.html" => "/policies",
+            "other-installers.html" => {
+                "https://forge.rust-lang.org/other-installation-methods.html"
+            }
+            "security.html" => "/policies/security",
+            "team.html" => "/governance",
```
Dmitry-Borodin: currently both 
https://www.rust-lang.org/en-US/team.html
https://www.rust-lang.org/en-US/governance
does not work
What should be used in the book?

nrc: The current URL is https://www.rust-lang.org/governance

nrc:
Travis failure is known Rustfmt/Rustup issue

RalfJung:
Did this get deployed yet? The links at https://github.com/rust-lang/www.rust-lang.org/issues/593#issuecomment-445395445 are still broken.

ashleygwilliams:
it is not deployed yet- i am backed up on reviews and need to get the rocket fix one in. it should be today. sorry about that!

PR #598, opened 2018-12-07
move myself to this account

komaeda:
moving github accounts, so i'm changing this


oe:
posting from this account to confirm that this is indeed me

LEARAX:
Why are you changing accounts?

PR #611, opened 2018-12-08
Add members of the embedded wg

jamesmunns:
This PR fills out the yaml of the current embedded wg members.

[Rendered](https://screenshots.firefox.com/2jBY8N2C3HyJqint/localhost)

As a note, the CSS seems to have trouble with oddly sized screens (using i3 with a lot of page zoom), but that's outside the scope of this PR.

[Issue Rendered](https://screenshots.firefox.com/Nhc0JZ0Oui3bNgu0/localhost)

### DIFF
--- a/src/data/wgs/embedded/team.yml
+++ b/src/data/wgs/embedded/team.yml
@@ -4,3 +4,31 @@ description: Focusing on the end-to-end experience of using Rust in resource-con
 chat: Domain Working Groups #wg-embedded channel on Discord
 repo: "https://github.com/rust-embedded/"
 leads: [japaric]
+members: [
+  adamgreig,
+  andre-richter,
+  awygle,
+  bradjc,
+  cr1901,
+  danc86,
+  dvc94ch,
+  dylanmckay,
+  Emilgardis,
+  hannobraun,
+  ithinuel,
+  jamesmunns,
+  jcsoo,
+  korken89,
+  nastevens,
+  nerdyvaishali,
+  paoloteti,
+  parched,
+  pftbest,
+  posborne,
+  raw-bin,
+  ryankurte,
+  sekineh,
+  thejpster,
+  therealprof,
+  wizofe,
+]


src/data/wgs/embedded/team.yml
```diff
@@ -4,3 +4,31 @@ description: Focusing on the end-to-end experience of using Rust in resource-con
...
+  paoloteti,
+  parched,
+  pftbest,
+  posborne,
+  raw-bin,
+  ryankurte,
+  sekineh,
+  thejpster,
+  therealprof,
+  wizofe,
```
Coder206: There is a trailing comma here, the other examples of members list omit the extra comma. Perhaps that why the build failed
```suggestion
  wizofe
```

jamesmunns:
Worth a try! If not there might be a `rocket_codegen` bug

jamesmunns:
Heu @ashleygwilliams, I just added a commit that checks in a `rust-toolchain` file rather than a `RustConfig` file, I think the travis ci might not be picking up the RustConfig correctly. We'll see, I can back out this change if preferred.

ashleygwilliams:
@jamesmunns i know the issue and it's with how the travis file adds rustfmt- could you back out those changes and just leave this as the team change? thanks!

jamesmunns:
@ashleygwilliams, okay, if you know what's up, that works for me. It seems that the `rust-toolchain` does work where the `RustConfig` didn't. I'm not familiar with `RustConfig`, though it seems like TravisCI is always pulling in the newest nightly, which doesn't tend to play nicely with Rocket.

I'll roll all the way back to 10c2b49be52027800dfce594269d857abf6a58d2 with a force push now.

jamesmunns:
@ashleygwilliams This is what made me suspect the missing `rust-toolchain` was the issue: https://travis-ci.org/rust-lang/www.rust-lang.org/builds/465217953#L757-L793

You can see that there are (likely) two issues:

1. Rocket Codegen fails to build (indicates breakage due to changing nightly)
2. rustfmt failed to install (probably the change you are referring to).

PR #614, opened 2018-12-08
fix #613

### LINKED ISSUE #613: cli "Get started!" code snippet variable mismatch and "?" operator invalid usage
# Content Bug Report

## Page(s) Affected
https://www.rust-lang.org/what/cli

## What needs to be fixed?
The Cli struct is defined with a "num" field, but in "main", the "take" function is having a "count" parameter passed to which does not exist.

The "?" operator can only be applied to values that implement "std::ops::Try", in the snippet it is applied to Cli  eg. "let args = Cli::from_args()?;"

## Suggested Improvement
In  the "main" function, replace "args.count" with "args.num"
Remove the "?" operator from the first line of the main function.

andreilorin:
Use correct parameter.
Remove illegal usage of "?" operator.

### DIFF
--- a/templates/what/cli/code-main.hbs
+++ b/templates/what/cli/code-main.hbs
@@ -2,10 +2,10 @@ use quicli::prelude::*;
 use structopt::StructOpt;
 
 fn main() -> CliResult {
-    let args = Cli::from_args()?;
+    let args = Cli::from_args();
     read_file(&args.file)?
         .lines()
-        .take(args.count)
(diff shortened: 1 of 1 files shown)


andreilorin:
Sorry I missed this the previous time..
main
[file.txt](https://github.com/rust-lang/www.rust-lang.org/files/2659668/file.txt)
cmd

nrc:
Thanks!

Issue #602, opened 2018-12-07
Website Fonts - TTF (404)


espnicholas:
# Technical Bug Report
## Your Environment
| Software         | Version(s) |
| ---------------- | ---------- |
| Form Factor      | All |
| Browser Type     | Chrome |
| Browser Version  | v71 |

## Is this an issue with Rust, CSS, or HTML (or a combination)?
HTML & Fonts - Specifically the /templates/layout.hbs file

## Expected Behavior
Trying to load .TTF fonts while I believe from another current issue they were removed and replaced with .WOFF. They are still referenced within the layout.hbs, thus trying to download and receiving a 404.

## Current Behavior
GET https://www.rust-lang.org/static/fonts/AlfaSlabOne-Regular.ttf net::ERR_ABORTED 404
15:00:35.395 www.rust-lang.org/:26 GET https://www.rust-lang.org/static/fonts/FiraSans-ExtraBold.ttf net::ERR_ABORTED 404
15:00:35.396 www.rust-lang.org/:27 GET https://www.rust-lang.org/static/fonts/FiraSans-ExtraBoldItalic.ttf net::ERR_ABORTED 404
15:00:35.396 www.rust-lang.org/:28 GET https://www.rust-lang.org/static/fonts/FiraSans-Italic.ttf net::ERR_ABORTED 404
15:00:35.398 www.rust-lang.org/:30 GET https://www.rust-lang.org/static/fonts/FiraSans-Regular.ttf net::ERR_ABORTED 404
15:00:35.399 www.rust-lang.org/:31 GET https://www.rust-lang.org/static/fonts/FiraSans-SemiBold.ttf net::ERR_ABORTED 404
15:00:35.399 www.rust-lang.org/:32 GET https://www.rust-lang.org/static/fonts/FiraSans-SemiBoldItalic.ttf net::ERR_ABORTED 404

## Screenshots
![screen shot 2018-12-07 at 3 04 26 pm](https://user-images.githubusercontent.com/2288231/49670158-b31a8000-fa31-11e8-886e-213d63804f32.png)
![screen shot 2018-12-07 at 3 07 01 pm](https://user-images.githubusercontent.com/2288231/49670174-c594b980-fa31-11e8-9429-ec8f8be4b065.png)

ashleygwilliams:
closed by #603

PR #626, opened 2018-12-10
Fixed path to static fonts

espnicholas:
I apologize,

I just realized that the static fonts were in a woff directory. This is to fix my last PR. #602 

Thanks!


chriskrycho:
@ashleygwilliams the Travis failure here is unrelated, so I'm going to go ahead and merge.

PR #631, opened 2018-12-11
Improved design of footer links

### LINKED ISSUE #610: Change :hover behaviour of yellow links in footer
The yellow links in the footer turn blue when hovered. This looks ugly in my opinion. Also, it is inherited from `skeleton.css`, so it's probably not intended or desired.

Instead, I suggest to make the links yellow all the time; also, I suggest to remove the **link underlines** and show them only when a link is hovered.

I'd also suggest to remove the circle bullets (only in the footer).

### Advantages

- The links still react to mouse interactions
- The links are better readable without the underline
- Less visual clutter
- The links don't change their colour

### Disadvantages

Everyone recognizes blue and/or underlined text as a link immediately. However, I believe that the visitors can easily figure out that the yellow text highlights are links.

 This is what it would look like:

issue links

Aloso:
Fixes #610

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -301,13 +301,25 @@ footer {
   padding: 30px 0;
   background-color: black;
   color: white;
+
   p {
     margin: 0;
   }
-}
 
-footer a {
-  color: $yellow;
+  ul {
+    list-style-type: none;
+    padding-left: 0;
+  }
+
+  a {
+    color: $yellow;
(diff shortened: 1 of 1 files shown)


nrc:
Could you add before and after screen shots please?

nrc:
Could you add before and after screen shots please?

Aloso:
Left side: before; right side: after

before_after

Issue #96, opened 2018-08-02
i18n / l10n method?


renyuneyun:
So, yeah, as it seems explicit the website needs to be i18n / l10n, there must be the backend support for that.
Presumbly this is done by using [fluent-rs](https://github.com/projectfluent/fluent-rs), and so the [Fluent Translation Language](https://projectfluent.org/fluent/guide/) will be the method of defining translation language.

I can imagine some behaviours which probably be good:

* Translations are placed in different directories (or simply different filenames) for different languages correspondingly
* Replace the text in the webpage template with variables (from `Context`), and load the variables dynamically (according to the language) in the rust code

There are two main questions before it can really be implemented into action:
1. Where to place the tranlation file?
    1. Should they be placed within the directory of the webpage template; or should they be placed together?
    2. Should the translations (in the same language) be placed in a hierarchical manner, or centralized?
2. How to name the translation text id?
    * When to use underscore `_` and when to use hyphen `-`?
    * Or, either way?

VitalyAnkh:
The new website will be shipped on Dec. 6 so how can we add i18n content now? I’m very glad to add Chinese translations but I don’t how to contribute.

skade:
Closing this, we have decided on a method.

PR #639, opened 2018-12-12
i18n scaffold

renyuneyun:
This is to fill the previous promise: #96  .

Eventually, it seems rust-fluent doesn't yet provide an easy-to-use method for production use. But either way, the scaffold for i18n shall be implemented, independent from fluent.

This is a work-in-progress version for that scaffold. All fluent-related parts are commented out.

Basically, a new file `i18n.rs` is added to handle it. It works as a fairing, by modifying the data proceeded by Handlebars templates.
In this implementation, any data in `::TEXT_ID::` are going to be translated: `TEXT_ID` extracted from here is fed to `i18n_text()` function, and the returned string will replace the original `::TEXT_ID::` part.

I'm wondering: 

1. How should the `shall_i18n()` function be working? The current one is really a hack, but I can't think of a better way at the moment.
2. Should I add a new trait to do the actual i18n work, and implement it? Do we need to bother doing so? (because it will never be a generic framework)

Hope rust-fluent can be utilized earlier, so that it can be used rather than having to "invent" a wrapper (not yet doing this, because the scaffold should be decided first).

(Changes will be made. I may force-push to the HEAD on my side...)

(I also modified the `Cargo.toml` because rocket v0.4 is in repo and the project won't compile in the previous `Cargo.toml`.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1700,6 +1700,7 @@ version = "0.1.0"
 dependencies = [
  "lazy_static 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.6.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "reqwest 0.9.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "rocket 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


renyuneyun:
So...

* A trait is added with some defaults (and those mod functions are removed).
* (Unnecessary) Comments are removed.

I think it's ready to take final revisions (if any from the team) and acts as a scaffold for i18n.

I'm now trying to integrate fluent into it, by implementing a `I18nProvider`.

PR #644, opened 2018-12-13
Show off serde in "POST some JSON" example

### LINKED ISSUE #482: "POST some JSON" example could show off serde
# Content Feature Request
## Summary
Instead of doing

```rust
let mut map = HashMap::new();
map.insert("lang", "rust");
map.insert("body", "json");
```

The example could try to show off serde by defining a struct similar to this:

```rust
#[derive(Serialize)]
struct Body {
    lang: String,
    body: String,
}
```

## Motivation
Serde is one of Rust's "killer apps": Not having to write (de)serialization code by hand is an amazing feature that I think deserves to be shown prominently on the website.

## Drawbacks
Using `String` for the field types means slightly nontrivial initialization using `"bla".to_string()` or `.into()`. Using `&'a str` means introducing lifetimes very early. Maybe this example should use integers or other types instead?

It makes the example longer: 2 additional lines for the struct definition + 2 when you create the struct inside the `json` method (4-5 if you create it out of line).

## Rationale and alternatives
- Why is this design the best in the space of possible designs?

  It shows how easy serialization is and what a more idiomatic way of performing this request looks like.
  
- What other designs have been considered and what is the rationale for not
  choosing them?

  The current example: `HashMap`s with stringly-typed keys/values are not generally considered best practice, so showing them off on the website might wrongly suggest that they are.

- What is the impact of not doing this?

  Stringly typed `HashMap`s are shown off as idiomatic code while they're generally not.

## Maintenance
Not needed anymore than with the existing example.

## Unresolved Questions
It's unclear which field types to use (see above).

jonas-schievink:
As an added bonus, this gets rid of the scroll bar in the code box
(at least on my machine).

Closes #482

### DIFF
--- a/templates/components/what/networking/post-json.hbs
+++ b/templates/components/what/networking/post-json.hbs
@@ -1,10 +1,15 @@
 // This will POST a body of
-//     `{"lang":"rust","body":"json"}`
-let mut map = HashMap::new();
-map.insert("lang", "rust");
-map.insert("body", "json");
+//     `{"delete": true, "user": 42}`
+#[derive(Serialize)]
+struct Body {
+    delete: bool,
+    user: u32,
+}
 
 let client = reqwest::Client::new();
-let res = client.post("http://httpbin.org/post")
-    .json(&map)
+let res = client.post("http://example.org")
+    .json(&Body {
+        delete: true,
+        user: 42,
+    })
     .send()?;
--- a/templates/components/what/networking/post-json.hbs
+++ b/templates/components/what/networking/post-json.hbs
@@ -1,5 +1,5 @@
 // This will POST a body of
-//     `{"delete": true, "user": 42}`
+//     `{"lang": "rust", "body": "json"}`
 #[derive(Serialize)]
 struct Body<'a> {
     lang: &'a str,
--- a/templates/components/what/networking/post-json.hbs
+++ b/templates/components/what/networking/post-json.hbs
@@ -1,15 +1,15 @@
 // This will POST a body of
 //     `{"delete": true, "user": 42}`
 #[derive(Serialize)]
-struct Body {
-    delete: bool,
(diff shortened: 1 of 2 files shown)


templates/components/what/networking/post-json.hbs
```diff
@@ -1,10 +1,15 @@
 // This will POST a body of
-//     `{"lang":"rust","body":"json"}`
-let mut map = HashMap::new();
-map.insert("lang", "rust");
-map.insert("body", "json");
+//     `{"delete": true, "user": 42}`
+#[derive(Serialize)]
+struct Body {
+    delete: bool,
```
spacekookie: I feel like keeping at least one string example in the struct could be useful. Strings are pretty ugly in most low-level languages and json deals with them a lot. Showing that Rust _doesn't_ have these issues could be good

jonas-schievink: I agree in principle, but using a string would require
* Introducing lifetimes very early (when using a `&'a str` as a struct field), or
* Making use of `.into()`/`.to_string()` conversion to get a `String`

(using `String` was the previous idea for this, see https://github.com/rust-lang/www.rust-lang.org/issues/482)

I think I'll let @aturon decide whether we should change. I don't really have a strong opinion here other than "we should totally use serde".

spacekookie: Hmm, yea. I see your points. Although considering that lifetimes are quite a central part of Rust, I'm not sure that we should hide their existence from code examples on the website *too much* either.

spacekookie:
I would want to mention strings, possibly with some lifetimes. They're not as scary as people think. And they're a core part of Rust.

The website should be advertising and showing off Rust, not hiding the code away :sweat_smile: 

ping @jonas-schievink mind making the small changes to your PR?

PR #623, opened 2018-12-10
align elements better by modernizing page to new rules

skade:
This ports the structure of the CLI page over to the embedded page, aligning buttons properly and giving consistent spacing.

Headlines still seem a bit fuzzy because of their length, but it's a clearly improved.

Before:
screenshot_2018-12-10 embedded - rust programming language 1

After: screenshot_2018-12-10 embedded - rust programming language

PR #649, opened 2018-12-14
Align Items on WASM Getting Started Page

### LINKED ISSUE #604: The Getting started Items of the WASM Page are misaligned
# Content Bug Report

## Page(s) Affected
The [wasm](https://www.rust-lang.org/what/wasm) page

## What needs to be fixed?
There seems to be a miss alignment of the getting started row
image

## Suggested Improvement
Fixing the alignment

Coder206:
Fixes #604 

Before: 
screenshot_2018-12-14 wasm - rust programming language 1

After:
screenshot_2018-12-14 wasm - rust programming language

Inspired by: #623

### DIFF
--- a/templates/components/what/wasm/get-started.hbs
+++ b/templates/components/what/wasm/get-started.hbs
@@ -4,32 +4,32 @@
             <h2>Get started!</h2>
             <div class="highlight"></div>
         </header>
-        <div class="row">
-            <div class="four columns">
+        <div class="flex flex-column flex-row-l">
+            <div class="flex flex-column justify-between-l mw8 measure-wide-l w-100 mt5 mt2-l">
                 <div class="domain-icon">
                   <img src="/static/images/webassembly-logo.svg" alt="WebAssembly Logo" />
                 </div>
-                <p>
+                <p class="flex-grow-1">
                     Learn more about the fast, safe, and open virtual machine called
                     WebAssembly, and read its standard.
                 </p>
                 <a href="https://webassembly.org/" target="_blank" rel="noopener" class="button button-secondary">Learn More</a>
             </div>
-            <div class="four columns">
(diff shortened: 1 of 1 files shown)


Coder206:
I would like to see the text aligned as well but I haven't had any success with using the baseline attributes

sondr3:
I think this SO answer might help: https://stackoverflow.com/a/39197187/273503. This is a bit of a limitation with Flexbox as far as I know, I think this would be fairly easy with CSS Grids (though I really have no idea).

Coder206:
Many thanks @sondr3!

r00ster91:
I think the images should become bigger again as well:
grafik
They are smaller than before.

Coder206:
Thank you for your patience everyone! Let me know what you think :)

yoshuawuyts:
@Coder206 Hmm, this is what it currently looks like on my machine:

## New
2019-01-18-174415_3840x2160

## Old
2019-01-18-174441_3840x2160

I feel this might not be what you were going for with this PR?

Coder206:
Agreed, just to be clear, are we looking to have the images vertically aligned, the first line of the text and the buttons at the same time?

yoshuawuyts:
> Agreed, just to be clear, are we looking to have the images vertically aligned, the first line of the text and the buttons at the same time?

@Coder206 definitely the top of the images and the buttons. I'm less sure about the text -- so probably?

Aligning text might require resizing images, which might take some effort. But if it's possible, perhaps it's worth a shot?

skade:
I think making sure that the images are all the same size is the way to go. @Coder206, is that a change you can do?

Coder206:
@skade I want to make sure I understand, you want me to make the widths of the images the same?

skade:
@Coder206 They should all fit the same box (same height and width), which might need some resizing of the images.

skade:
Fixed by skeleton removal, https://github.com/rust-lang/www.rust-lang.org/issues/830 is a followup.

PR #652, opened 2018-12-14
Cache the current Rust version.

mbrubeck:
Currently, every time the home page is rendered, the web server downloads a TOML file from static.rust-lang.org and parses it to find the current Rust version.  This adds a significant amount of latency to the home page.

This patch caches the version in memory for up to 10 minutes.  This greatly speeds up home page rendering, at the cost of a 10-minute delay before a new version is reflected on the home page.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1683,6 +1683,7 @@ dependencies = [
 name = "wubwub"
 version = "0.1.0"
 dependencies = [
+ "lazy_static 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum:
This seems good to me -- however, a ten minute delay might be a bit unfortunate. Could we perhaps lower that down to around 2 minutes? Ideally we'd also change the code such that the cache refresh happens in the background, i.e., we reply immediately and enqueue the refresh onto a separate thread. That might not be feasible on Heroku (or really all that necessary, though).

Mark-Simulacrum:
Thanks!

PR #654, opened 2018-12-14
Add some padding in the domains block

### LINKED ISSUE #543: Add margin between build it in rust intro and the domains section
# Design Feature Request
## Summary
On landing page, I think we should add more margin between the phrase after "build it en Rust" and the next section.
screenshot_20181204_215638

The spacing seems odd, unbalanced.

My proposition would be :

screenshot_20181204_220122

(Adding 25px more to the phrase block margin bottom)

I can volunteer to send a PR.

meven:
Fix #543 
Before
before
After
after

I am myself not a fan of inline styling but I would need some guidance as to where to put it elsewhere.

### DIFF
--- a/templates/components/what/cli/maintainable.hbs
+++ b/templates/components/what/cli/maintainable.hbs
@@ -7,7 +7,7 @@
         <div class="row">
             <article class="six columns box">
                 <header class="domain-icon">
-                    <img src="/static/images/cli-app-structure-2.svg" alt="" />
+                    <img src="/static/images/cli-app-structure-2.svg" alt="cli app structure" />
                   <h3>Catch errors <em>now</em></h3>
                 </header>
                 <p class="box-content">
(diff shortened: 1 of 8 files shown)


templates/components/panels/domains.hbs
```diff
@@ -8,7 +8,7 @@
     </header>
 
     <div class="flex-none flex-l flex-row">
-      <p class="flex-grow-1">
+      <p class="flex-grow-1" style="padding-bottom:25px">
```
spacekookie: Since I recently found that having some kind of padding attribute in the CSS in general for various sections, I'm wondering if this could be a good thing to put into a very generic `padding` class (for various directions)?

yoshuawuyts:
```suggestion
       <p class="flex-grow-1 pb2">
```

`pb2` or `pb3` might perhaps work better. Alternatively the next section could have a `mt2` or `mt3` which is generally more consistent with how styles have been applied throughout the rest of the site.

PR #668, opened 2018-12-25
Fix unclosed element div

spk:
This fix markup error with unclosed div element (ref https://validator.nu/?doc=https%3A%2F%2Fwww.rust-lang.org%2F)
Cheers,
Laurent

### DIFF
--- a/templates/components/panels/get-involved.hbs
+++ b/templates/components/panels/get-involved.hbs
@@ -4,28 +4,28 @@
       <h2>Get involved</h2>
       <div class="highlight"></div>
     </header>
-      <div class="row media">
-        <div class="six columns" id="read-rust">
-          <h3>Read Rust</h3>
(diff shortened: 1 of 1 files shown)


templates/components/panels/get-involved.hbs
```diff
@@ -2,7 +2,7 @@
   <div class="w-100 mw-none ph3 mw8-m mw9-l center f3">
     <header>
       <h2>Get involved</h2>
-      <div class="highlight">
+      <div class="highlight"></div>
     </header>
       <div class="row media">
```
yoshuawuyts: Hm, we should probably change the indentation here too to prevent any future regressions. Would you mind moving this section one to the left so the indentation lines up again?

spk: Yep done in https://github.com/rust-lang/www.rust-lang.org/pull/668/commits/1b1187618dc3dccfea20b8046c832a28d2811e61

PR #680, opened 2019-01-01
Add/update some img alts

spk:
Hi this add/update some image alternative's, cheers

### DIFF
--- a/templates/components/what/networking/production.hbs
+++ b/templates/components/what/networking/production.hbs
@@ -41,7 +41,7 @@
             </div>
             <div class="testimonial row">
                 <div class="four columns">
-                    <img src="/static/images/1aim.svg" alt="laim" />
+                    <img src="/static/images/1aim.svg" alt="1aim" />
                 </div>
                 <div class="eight columns">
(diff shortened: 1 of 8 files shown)


templates/components/what/networking/production.hbs
```diff
@@ -36,12 +36,12 @@
...
                 </div>
                 <div class="four columns">
-                    <img src="/static/images/buoyant.svg" />
+                    <img src="/static/images/buoyant.svg" alt="buoyant" />
                 </div>
             </div>
             <div class="testimonial row">
                 <div class="four columns">
-                    <img src="/static/images/1aim.svg" />
+                    <img src="/static/images/1aim.svg" alt="laim" />
```
yoshuawuyts:
```suggestion
                     <img src="/static/images/1aim.svg" alt="1aim" />
```

spk: Opps nice catch I did not see the diff with my fonts thanks, fixed in  c44b973f57ce48884a757ee46cb42fc1c0d417bf

PR #681, opened 2019-01-02
Use 127.0.0.1 instead of localhost for dev to avoid lookup

spk:
Hi just using ` 127.0.0.1 ` in Rocket development config instead of `localhost` to avoid lookup; cheers

### DIFF
--- a/Rocket.toml
+++ b/Rocket.toml
@@ -1,3 +1,3 @@
 [development]
-address = "localhost"
+address = "127.0.0.1"
 port = 7878


yoshuawuyts:
@spk could you say more about what this does? Which lookup is removed?

spk:
Hi @yoshuawuyts, this just avoid the server to resolv every boot the name `localhost` to the IP it need to listen to, sorry if my first comment was not clear

yoshuawuyts:
@spk all good! -- thanks for the contribution!

Issue #655, opened 2018-12-15
Chucklefish logo too large


Maxgy:
The Chucklefish logo on the main page seems to be just a little too large:

![image](https://user-images.githubusercontent.com/19353212/50046713-fc3b8700-0064-11e9-90dc-4ab66977f6b5.png)

mglax:
Would be cool to make the image responsive too. Here is how it renders on an iPhone 6
![3af29740-c8e0-43f0-9fa7-957141758fb7](https://user-images.githubusercontent.com/44877814/50378812-30054680-063b-11e9-876c-74b1b6f59eac.jpeg)

I think it's a bit too large.

Maxgy:
closing this issue since the Chucklefish logo is no longer present

PR #689, opened 2019-01-18
Fixing chucklefish logo size (Issue #655)

Anish-Agnihotri:
**Intro**

Hey folks! Noticed that #655 was significant and had some discussion about potential size reductions. I've made this quick PR that adds a one-liner html id, and a few lines of responsive CSS to make sure that the chucklefish logo is a better fit at all screen sizes.

Looking to help out with some of the open issues, and went through the Contributors guidelines, but do let me know if I'm out of place!

Thanks,

Anish

**New sizing screenshots**

The new sizing should look better on smaller displays and be less obtrusive on larger ones. It is responsively sized based on width: percentage.

Large display:
large_display

Medium display:
medium_size

Mobile display: 
mobile_reduction

**Resolves**
#655


Anish-Agnihotri:
On the same note, would it be a better idea to keep things simple and simply resize the source PNG?

templates/components/panels/production.hbs
```diff
@@ -22,7 +22,7 @@
           <p class="attribution">&ndash; Catherine West, Technical Lead</p>
         </div>
         <div class="four columns">
-          <a href="https://blog.chucklefish.org/about/" target="_blank" rel="noopener"><img src="/static/images/user-logos/chucklefish.png" alt="Chucklefish Logo" /></a>
+          <a href="https://blog.chucklefish.org/about/" target="_blank" rel="noopener"><img src="/static/images/user-logos/chucklefish.png" alt="Chucklefish Logo" id="chucklefish" /></a>
```
yoshuawuyts: We could perhaps express this with less code using Tachyons:
```suggestion
          <a href="https://blog.chucklefish.org/about/" target="_blank" rel="noopener"><img src="/static/images/user-logos/chucklefish.png" alt="Chucklefish Logo" class="w-33 w-50-ns" /></a>
```

Anish-Agnihotri: Make complete sense. My apologies, I'll pay more attention to simple tachyon-backed solutions.

Just pushed another commit that removes the class based width style: `width: 80%` and instead gives the images a tachyons `w-` attribute.

yoshuawuyts:
> On the same note, would it be a better idea to keep things simple and simply resize the source PNG?

I think the approach you've taken is really good, and the results look great! :tada:

PR #690, opened 2019-01-18
Update dev-tools teams

nrc:


src/data/teams/dev-tools/team.yml
```diff
@@ -3,5 +3,5 @@ name: Dev Tools team
 description: Rust developer tools
 chat: "Dev Tools #dev-tools channel on Discord"
 repo: "https://github.com/rust-dev-tools/dev-tools-team"
-leads: [nrc]
-members: [fitzgen, japaric, killercup, manishearth, steveklabnik, withoutboats]
```
yoshuawuyts: Heh, it might be odd if there's nobody left in this team. Does that mean it's effectively disbanded, or should we link people to the relevant subteams perhaps?

nrc: It is temporarily disbanded, but acting as a way to group the sub-teams and WGs

Issue #691, opened 2019-01-19
Link to blog post for latest release


dustinknopoff:
# Content Feature Request
## Summary
https://internals.rust-lang.org/t/release-blog-post-link-on-website-front-page/9074 

r? @Mark-Simulacrum

PR #692, opened 2019-01-19
[WIP] latest release blog post

dustinknopoff:
#691 

r? @Mark-Simulacrum

### DIFF
--- a/src/rust_version.rs
+++ b/src/rust_version.rs
@@ -1,4 +1,5 @@
 use reqwest;
+use serde_json;
 use std::sync::RwLock;
 use std::thread;
 use std::time::Instant;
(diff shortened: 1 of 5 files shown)


dustinknopoff:
I'm unsure where to go from here. Should I be creating a cache for the latest release? Should it be stored as a string or in a struct?

Mark-Simulacrum:
We'll probably want to fetch the releases on the same schedule as new Rust versions (probably even in the same function), also caching them - I would even do so in the same static.

I think for now we only need the URL from the post and we should be able to directly return an `a` html tag.

dustinknopoff:
I think the functions could use better names but it works, and is added as the link location for `Version 1.32.0` under the `Get Started` button.

Mark-Simulacrum:
It looks like with the change from `<p>` to `<a>` it's no longer centering the version:

image

Could you maybe try moving the anchor into the (previous) paragraph? Something like `<p><a>...</a></p>`?

Mark-Simulacrum:
Let's squash the commits here and this looks ready to merge!

src/rust_version.rs
```diff
@@ -37,3 +53,16 @@ fn fetch_rust_version() -> Option<String> {
...
     Some(version)
 }
+
+fn fetch_rust_release_post() -> Option<String> {
+    let releases = reqwest::get("https://blog.rust-lang.org/releases.json")
+        .ok()?
+        .text()
+        .ok()?;
+    let releases = releases.parse::<serde_json::Value>().ok()?;
+    let url = releases["releases"][0]["url"].as_str()?.to_string();
```
dustinknopoff: Just realized. Should this be `releases["releases"][0]`?

Mark-Simulacrum: Not sure I follow you - and local testing indicated that the site works and correctly links.

dustinknopoff: It works because there is only one release right now. But I'm uncertain the order if there are multiple releases. Should it be the first element or the last?

Mark-Simulacrum: Ah. Hm. I believe it'll be the first post if I understand the blog code correctly so we should be fine; if it doesn't work then changing it is fairly easy so I'm not too worried.

PR #693, opened 2019-01-20
Random refactorings

pietroalbini:
This PR does two refactorings over the codebase:

* Use a single `Context` struct containing all the data needed in the templates, to avoid duplication
* Extract the caching logic from `rust_version.rs` to its own module, and make it generic over the accepted data. This allows other parts of the website to cache stuff without reimplementing everything. I also added a few tests to make sure the cache works.

These refactorings are needed for a PR I'm working on to fetch the team structure from the new, WIP team structure repository ([rust-lang/team](https://github.com/rust-lang/team), private at the moment).


src/cache.rs
```diff
@@ -0,0 +1,152 @@
...
+        .map(|&(ref data, timestamp)| {
+            if timestamp.elapsed().as_secs() > CACHE_TTL_SECS {
+                // Update the cache in the background
+                thread::spawn(move || {
+                    let _ = update_cache::<T>(generator);
+                });
+            }
+            data
+        })
+        .and_then(|data| data.downcast_ref::<T>().map(|t| t.clone()))
```
Mark-Simulacrum: I think unwrap/clone would make more sense here -- if this isn't a T that's a bug in the code, not a missing value, correct?

src/cache.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+fn update_cache<T>(generator: Generator) -> Result<T, Box<Error>>
+where
+    T: Send + Sync + Clone + 'static,
+{
+    if let Ok(data) = generator()?.downcast::<T>() {
```
Mark-Simulacrum: Would it perhaps make more sense for the generator to return `Result<T, Box<Error>>` and for us to wrap the T in a Box<Any> here?

src/rust_version.rs
```diff
@@ -1,68 +1,44 @@
...
-    let cached = CACHE.read().unwrap();
-    let (version, timestamp) = cached.as_ref()?;
-    if timestamp.elapsed().as_secs() > CACHE_TTL_SECS {
-        // Update the cache in the background.
-        thread::spawn(fetch_rust_version);
+pub fn rust_release_post() -> Option<String> {
+    match ::cache::get(fetch_rust_release_post) {
+        Ok(post) => Some(post),
+        Err(err) => {
+            println!("error while fetching the rust release post: {}", err);
```
Mark-Simulacrum: I think these two println's should probably go to stderr, not stdout.

src/cache.rs
pietroalbini: Yes, I'll use unwrap.

src/cache.rs
pietroalbini: Don't think that would work, since the function pointer would then have to be generic over `T` and we wouldn't be able to use it as the hashmap key.

pietroalbini:
Addressed review comments.

PR #702, opened 2019-02-06
Reject invalid governance pages and add some tests

pietroalbini:
This PR rejects standalone governance pages for subteams, and adds a few tests on the backend code to make sure we don't regress on that page.

### DIFF
--- a/src/teams.rs
+++ b/src/teams.rs
@@ -1,6 +1,7 @@
 use reqwest;
 use rust_team_data::v1::{Team, TeamKind, Teams, BASE_URL};
 use std::any::Any;
+use std::cmp::Reverse;
 use std::error::Error;
 use std::fmt;
 
@@ -24,6 +25,7 @@ pub struct PageData {
     wgs: Vec<Team>,
 }
 
+#[derive(Clone)]
 struct Data {
     teams: Vec<Team>,
 }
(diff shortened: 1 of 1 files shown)


src/teams.rs
```diff
@@ -4,6 +4,113 @@ use std::any::Any;
...
+            })
+            .for_each(|team| match team.team.kind {
+                TeamKind::Team => data.teams.push(team),
+                TeamKind::WorkingGroup => data.wgs.push(team),
+            });
+
+        data.teams
+            .sort_by_key(|team| -team.team.website_data.as_ref().unwrap().weight);
+        data.wgs
+            .sort_by_key(|team| -team.team.website_data.as_ref().unwrap().weight);
```
Mark-Simulacrum: Could we perhaps utilize `cmp::Reverse` here instead?

src/teams.rs
```diff
@@ -4,6 +4,113 @@ use std::any::Any;
...
+                ),
+                team,
+            })
+            .for_each(|team| match team.team.kind {
+                TeamKind::Team => data.teams.push(team),
+                TeamKind::WorkingGroup => data.wgs.push(team),
+            });
+
+        data.teams
+            .sort_by_key(|team| -team.team.website_data.as_ref().unwrap().weight);
```
Mark-Simulacrum: I personally would like this to be replaced with e.g. the below or otherwise made to not read as `team.team` which just looks quite odd with relatively little gain I think.

```suggestion
            .sort_by_key(|idx_team| -idx_team.team.website_data.as_ref().unwrap().weight);
```

pietroalbini:
Hopefully fixed all the issues.

pietroalbini:
Ping @Mark-Simulacrum @yoshuawuyts

Mark-Simulacrum:
Feel free to merge when ready - I won't be around for the next few hours so don't want the website to go down without someone to fix it.

PR #709, opened 2019-02-14
add rustlings to the learn section

komaeda:
#### Preview

image

/cc @steveklabnik


templates/learn/index.hbs
```diff
@@ -20,14 +20,23 @@
           you’ll have a solid grasp of the language.</p>
           <a class="button button-secondary" href="https://doc.rust-lang.org/book/">Read the Book!</a>
       </section>
-      <section class="pl5-l pt4 pt0-l">
+    </div>
+    <div class="flex flex-column flex-row-l justify-between">
+      <section class="pt0">
```
yoshuawuyts: Perhaps a slightly nicer way of modelling this would be to follow what we did here: https://github.com/rust-lang/www.rust-lang.org/blob/9a9f1cfeced812d084040ad55ab63ab7254efe41/templates/learn/index.hbs#L140-L196

This puts the buttons on the same height, regardless of the text size, making the flow a bit calmer.

yoshuawuyts: What it looks like: 

2019-02-19-191749_1920x1080

komaeda: good idea, i'll try this today!

komaeda: @yoshuawuyts i've tried modelling it after this, let me know what you think!

Issue #726, opened 2019-03-12
Are there any solutions to view other platform installation page?


lzutao:
# Question

## Question

This maybe related to #423.

I'm trying to find a link that talk about **two different toolchains** on 
Windows (see [rust-lang/rustup.rs#1707#discussion_r264291131](https://github.com/rust-lang/rustup.rs/pull/1707#discussion_r264291131)). 
The old link https://www.rust-lang.org/downloads.html#win-foot is no longer exists.

Can I view other platform's installation page (https://www.rust-lang.org/tools/install) 
with a URL HTML marker (`#platform-win`) or URL GET paramenter (`?platform=windows`) ?

## Context

I'm using Ubuntu and I want to view Windows installation page.

After skimming through install page of r-l.o code, it seems like `platform-button` 
[is not used in `install-script.js`](
https://github.com/rust-lang/www.rust-lang.org/blob/master/templates/components/tools/install-script.hbs#L114-L141).

yoshuawuyts:
@lzutao what screen are you referring to? Is that https://www.rust-lang.org/learn/get-started ?

lzutao:
@yoshuawuyts Sorry for ambiguous. I've updated the issue with more context.

yoshuawuyts:
@lzutao hah, yeah this has annoyed me before too. I think having a query string indeed makes for a reasonable (interim) solution that can be added without too much trouble. Would you be willing to write a patch to add it?

lzutao:
@yoshuawuyts kinnison has written a patch for this issue. Feel free to take a look at it.

PR #729, opened 2019-03-16
Tweak installation page

kinnison:
In order that we can link nicely to the installation page for various platforms from the `rustup.rs` documentation (for example, in order to support https://github.com/rust-lang/rustup.rs/pull/1707#discussion_r264291131 and similar) this PR offers up some tweaks to the installation page to do just that.

It is split into three commits.  The first just tweaks the formatting and wording of the "generic" or "default" instructions a little, the second fixes a JS error when the test button is absent from the install page, and the third introduces the query-string based override support.  In theory only the third is needed, but I think the other two are worthwhile tweaks also.

### DIFF
--- a/templates/components/tools/install-script.hbs
+++ b/templates/components/tools/install-script.hbs
@@ -113,7 +113,9 @@ function cycle_platform() {
 
 function set_up_cycle_button() {
     var cycle_button = document.getElementById("platform-button");
-    cycle_button.onclick = cycle_platform;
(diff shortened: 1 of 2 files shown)


kinnison:
Since Github hasn't spotted the link, this should fix #726

Mark-Simulacrum:
Seems good to me but I'm personally not ready to just merge this, would like a review from @yoshuawuyts as well. I have not tested locally.

PR #731, opened 2019-03-26
[WIP] I18n fluent implementation

renyuneyun:
So, this is a (large?) step towards fluent implementation of the i18n mechanism, as mentioned in (and built on top of) #639 .
(Also see #732 for a variant of this approach.)

Basically, there is one problem remaining, related to lifetime. The rest (fluent, reading) should work fine, though apparently code clean up and refactor is needed (but that's after "working", right?). It assumes [FTL](http://projectfluent.org/fluent/guide/) being stored under `templates/fluent-resource/LANGUAGE-CODE`.

As you can see, the compiler says the lifetime is not long enough.

* It seems rocket requires the fairing to contain only static references.
* However, the construction of `FluentBundle` (which will be passed to the fairing) requires references to strings.
  * These strings can only be read during runtime, because they are the FTL stored in files.

So, how should these strings be stored, in order to satisfy the requires?

### DIFF
--- a/src/fluent_wrapper.rs
+++ b/src/fluent_wrapper.rs
@@ -1,0 +1,86 @@
+use std::collections::HashMap;
+use std::error::Error;
+use std::fs::File;
+use std::fs::read_dir;
+use std::io::BufReader;
+use std::io::prelude::*;
+use std::io::Result;
+use std::path::Path;
+
+use fluent_bundle::{FluentBundle, FluentResource};
+
+use i18n::I18NProvider;
+
+pub fn read_from_file<P: AsRef<Path>>(filename: P) -> Result<Vec<FluentResource>> {
+    let mut file = try!(File::open(filename));
+    let reader = BufReader::new(file);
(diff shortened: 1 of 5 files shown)


src/fluent_wrapper.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use fluent_bundle::{FluentBundle, FluentResource};
+
+use i18n::I18NProvider;
+
+pub fn read_from_file<P: AsRef<Path>>(filename: P) -> Result<Vec<FluentResource>> {
+    let mut file = try!(File::open(filename));
+    let reader = BufReader::new(file);
+
+    let mut fluent_resources = Vec::new();
```
renyuneyun: This part is unnecessary because `FluentResource` can handle multiple pairs. Only a full read of the file is needed, and the whole string passed to construct `FluentResource`.

skade:
Merged through #790 \o/

PR #732, opened 2019-03-27
I18n fluent implementation2

renyuneyun:
A slight different approach to #731 , changed where the `FluentBundle` is initialized: from `main()` to `FluentI18nProvider::new()`.

However, the problem is the same: lifetime.
I also tried to store `FluentResource`s in `FluentI18nProvider`, but it still says lifetime problem. Even if I tried `Arc`.

Potentially, [this SO post](https://stackoverflow.com/questions/32300132/why-cant-i-store-a-value-and-a-reference-to-that-value-in-the-same-struct) and [this](https://stackoverflow.com/questions/30538387/store-reference-of-struct-in-other-struct) are related to the problem here?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,20 +7,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "aho-corasick"
-version = "0.6.9"
+version = "0.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 2.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
-name = "arrayref"
-version = "0.3.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
 name = "arrayvec"
-version = "0.4.8"
+version = "0.4.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "nodrop 0.1.13 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -31,9 +26,35 @@ name = "atty"
 version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.45 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.54 (registry+https://github.com/rust-lang/crates.io-index)",
  "termion 1.5.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "winapi 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "autocfg"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "backtrace"
+version = "0.3.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "autocfg 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "backtrace-sys 0.1.28 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cfg-if 0.1.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.54 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rustc-demangle 0.1.14 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "backtrace-sys"
+version = "0.1.28"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "cc 1.0.37 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.54 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -41,22 +62,17 @@ name = "base64"
 version = "0.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "byteorder 1.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "byteorder 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "safemem 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "base64"
-version = "0.10.0"
+version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "byteorder 1.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "byteorder 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
-
-[[package]]
-name = "bitflags"
-version = "0.7.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "bitflags"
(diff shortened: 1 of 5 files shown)


skade:
@renyuneyun I can't see this error in the build log here, it mainly complains about wrong use statements.

skade:
@renyuneyun does this PR supersede the other two PRs? #639 and #731 ?

renyuneyun:
Hi @skade 
The three PRs are like this:

* #639 is the original and basic (scaffold) for the i18n effort, and it does not introduce any external dependencies. It's meant to be generic.
* #731 and #732 (this PR) are two efforts for trying to make rust-fluent work. They use slightly different approaches and each has a different problem.

So both #731 and #732 supersede #639; only one of #731 and #732 is needed in the end if merged.


I'll look into the build error when I switch back to my machine. I assume it's because of stable and nightly version?

renyuneyun:
A little more information for the status: I was waiting for the upstream (projectfluent/fluent-rs#64) previously, to solve the problem I encountered when trying to make it work (i.e. this PR and #731).

There is some new comments on the upstream, so I'll try to see if I can make it work when I have time (likely a weekend soon).

skade:
Okay, so I'll close #639, as we're pretty decided to use fluent.

Thanks for clarifying and working on this :).

Manishearth:
That issue should be fixed for us now, Fluent 1.0 has a resource manager that's useful for this use case.

Manishearth:
Playing around with this we'll need to upgrade the compiler used here to be able to use an updated Fluent.

Manishearth:
I chatted with Zibi about this: currently, the built in Fluent ResourceManager is single threaded: you create it in your `main()` and pass it down to everyone.

Fluent is designed so that you can create your own resource manager: The resource manager is basically just a cache.

We have a couple options here:

 - Not use a resource manager at all, and just create bundles on the fly. This is expensive; we should only do this if Rocket can cache the generated pages itself (can it? looks like you need additional crates for this)
 - Create our own resource manager based on `elsa::sync::FrozenMap` (Fluent uses `elsa::FrozenMap`). I'd need to land https://github.com/Manishearth/elsa/pull/1 for this, though.

Thoughts?

renyuneyun:
@skade I rebased the code to the current master, and you should be able to see the error.

Basically it complains about lifetime. It's about the `fluent_resources` at line 72 of file `fluent_wrapper.rs`.
The solution I can think of is to store this variable in the struct. Something like this:
```
pub struct FluentI18nProvider<'a> {
    fluent_resource_collection: FluentResourceCollection,
    collection: HashMap<String, FluentBundle<'a>>,
}

impl <'a> FluentI18nProvider<'a> {
    pub fn new() -> FluentI18nProvider<'a> {
        ...
        FluentI18nProvider {
            fluent_resource_collection: fluent_resources,
            collection: fluent_collection,
        }
    }
}
```

But if I do this, the compiler would complain
```
error[E0515]: cannot return value referencing local data `*fluent_resources`
  --> src/fluent_wrapper.rs:89:9
   |
84 |           for (lang, resources) in &*fluent_resources {
   |                                    ------------------ `*fluent_resources` is borrowed here
...
89 | /         FluentI18nProvider {
90 | |             fluent_resource_collection: fluent_resources,
91 | |             collection: fluent_collection,
92 | |         }
   | |_________^ returns a value referencing data owned by the current function

error[E0505]: cannot move out of `fluent_resources` because it is borrowed
  --> src/fluent_wrapper.rs:89:41
   |
71 |   impl <'a> FluentI18nProvider<'a> {
   |         -- lifetime `'a` defined here
...
83 |           for (lang, resources) in &fluent_resources {
   |                                    ----------------- borrow of `fluent_resources` occurs here
...
88 | /         FluentI18nProvider {
89 | |             fluent_resource_collection: fluent_resources,
   | |                                         ^^^^^^^^^^^^^^^^ move out of `fluent_resources` occurs here
90 | |             collection: fluent_collection,
91 | |         }
   | |_________- returning this value requires that `fluent_resources` is borrowed for `'a`
```

Then, I have no idea how to solve this... I was expecting that `FluentI18nProvider {fluent_resource_collection: fluent_resources, collection: fluent_collection,}` *moves* `fluent_resources`, so that the reference is kept available. But it seems not, at least the compiler doesn't agree.

Then I thought it's because `HashMap` is not allocated in the heap, so that *move* doesn't really apply? But even if I wrap it with `Box<FluentResourceCollection>`, there is no real different...

So I'm really confused... Any hints?

renyuneyun:
@Manishearth I'm ok with both.

I tried fluent-resource-manager just now. My effort is to change the `FluentI18nProvider` to `pub struct FluentI18nProvider {resmgr: ResourceManager}`, and change the `new()` function.

It turned out that it too needs Manishearth/elsa#1, as the compiler complains about `FluentI18nProvider` can not be shared across threads (because of `elsa::sync::FrozenMap` eventually).

But from the API of fluent-resource-manager, I would say a separate resource manager is slightly better.

Manishearth:
I'll land and publish that, then

Manishearth:
Published elsa 1.2.

What you want to do is something like https://github.com/zbraniecki/fluent-rs/blob/89495dff62d47a2d3de22d6a23fe8c19518e2513/fluent-resmgr/src/resource_manager.rs#L36 , except using the sync FrozenMap.

Make sure you use an updated Fluent, and rebase over master.

Manishearth:
Oh, and you can stick the sync ResourceManager into a global static or something.

Manishearth:
Oh, no, there are still problems. THough there is one way to fix this that doesn't require elsa at all. Bit inefficient, but it's okay.

Manishearth:
I got it working, elsa wasn't necessary at all. See https://github.com/Manishearth/www.rust-lang.org/tree/fluent

I also fixed a bug you had: your implementation was parsing individual file lines as fluent resources, we should be parsing entire files as resources

renyuneyun:
@Manishearth Thanks for the reference. I have made the adjustments with other improvements.

It is now a functioning version, with a very simple fallback: to `en`. It works like this:

1. Create language-specific resource directory under `templates/fluent-resource/`, e.g. `templates/fluent-resource/en/` (the path `templates/fluent-resource` is passed as an argument)
2. Put translation resource files under the directory (in any names), including the default texts (under `/en/`)
3. Replace text blocks in the HTML templates with `::TEXT_ID::` to refer to a fluent resource with id `TEXT_ID`
4. Visit the website and pass the desired language as query with key`lang`, e.g. `localhost:7878/?lang=zh` for language `zh`; `en` is the default value is no language is specified

It doesn't handle any advanced uses of fluent, e.g. passing external variables, as mentioned in [the guide of fluent](https://projectfluent.org/fluent/guide/placeables.html). But it should be fairly enough for static translations.

renyuneyun:
Yet, I have no idea why the current build error exists, when it comes to rust2018...
It looks the same as other local modules, but why is it...?

Manishearth:
You're doing it from a submodule. Just change it to `crate::`

Manishearth:
Should we be landing this, then? I can start reviewing if you'd like.

I'd love to see something that starts adding ids to be translated.

renyuneyun:
@Manishearth Thanks. It works like a charm.
But still, why is only this `i18n` module affected? I mean, other many modules are also submodules, but the compiler only complains about `i18n`.

Is it because `i18n` has been used in more than one places (`main.rs` & `fluent_wrapper.rs`)?

Manishearth:
> I mean, other many modules are also submodules, but the compiler only complains about i18n.

No, I mean you're importing it from a module.

The imports work within `main.rs` because the submodules are a part of the crate root, and you're at the crate root. In `fluent_wrapper.rs` you're in a submodule already, and `i18n` is not a part of the submodule, so you have to explicitly say that you want a module from the crate root.

renyuneyun:
Yes, I'd like to see it merged and to see a localized rust website :)

(But I may only be able to deal with the comments tomorrow. It's almost 24:00 where I am.)

renyuneyun:
@Manishearth Thanks a lot. I didn't notice the compiler is talking about `fluent_wrapper.rs` but not `main.rs`... I may need to rest and clear my brain haha.

skade:
Had to do some fixup, I'll merge once CI runs through.

PR #735, opened 2019-03-30
Add missing arg for new project generation example

kevincon:
It seems `cargo new` currently defaults to using a library template to create a new project, but the "Generating new project" example assumes it will use the binary template. 

The library template is incompatible with the next steps in the example. When you try to run `cargo run`, you get an error:

```
➜  hello-rust git:(master) ✗ cargo run
error: a bin target must be available for `cargo run`
```

This PR adds the missing `--bin` argument to make the example correct.

### DIFF
--- a/templates/learn/get-started.hbs
+++ b/templates/learn/get-started.hbs
@@ -58,7 +58,7 @@
     </header>
     <p>Let’s write a small application with our new Rust development environment. To start, we’ll use Cargo to
       make a new project for us. In your terminal of choice run:</p>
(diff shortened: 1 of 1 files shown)


kevincon:
Ah nevermind, I had just installed rustup but it seems it gave me an ancient version of `cargo` (`cargo 0.24.0 (45043115c 2017-12-05)`) where the library template was the default. After running `rustup update` I got version `cargo 1.33.0 (f099fe94b 2019-02-12)` and now from `cargo help new` I see the binary template is the default. :)

Issue #720, opened 2019-03-03
Remove Chucklefish Whitepaper


jonathandturner:
# Content Bug Report

## Page(s) Affected
Chucklefish is not using Rust in Witchbrook anymore, after Catherine West left. 

Source(s):
https://www.reddit.com/r/programming/comments/atyzz4/halley_a_lightweight_game_engine_written_in_c14/eh69lz7/

https://www.reddit.com/r/rust/comments/avwxq1/chucklefish_is_no_longer_using_rust_for_witchbrook/ehijdwx/

## What needs to be fixed?
Remove the Chucklefish PDF and possibly the quote related to Chucklefish.

josh-kean:
hi jonathandturner, I found the page that Chucklefish is mentioned on, and I think I can effectivley remove it. Mind if I take a crack at this?

lzutao:
@josh-kean You're welcome to do that.

skade:
Hm, there's a PR for that, which was closed.

PR #740, opened 2019-04-13
removed chucklefish white paper

josh-kean:
This commit would remove the chucklefish whitepaper as brought up in issue #720.


skade:
@josh-kean sorry for the delay, i took your PR, fixed it a little and merged it through #771. Thanks for the work!

PR #741, opened 2019-04-16
add `extern crate ferris_says` to get started example

matt-h:
add `extern crate ferris_says` to get started example

https://www.rust-lang.org/learn/get-started

At least with rust version 1.34 the existing example errors with the following error:

```
error[E0432]: unresolved import `ferris_says`
 --> src/main.rs:1:5
  |
1 | use ferris_says::say; // from the previous step
  |     ^^^^^^^^^^^ maybe a missing `extern crate ferris_says;`?

error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.
error: Could not compile `hello-rust`.

To learn more, run the command again with --verbose.
```

### DIFF
--- a/templates/learn/get-started.hbs
+++ b/templates/learn/get-started.hbs
@@ -112,7 +112,8 @@ ferris-says = "0.1"</code></pre>
     </header>
     <p>Now let’s write a small application with our new dependency. In our <code>main.rs</code>, add the following
       code:</p>
-    <pre><code>use ferris_says::say; // from the previous step
+    <pre><code>extern crate ferris_says;
+use ferris_says::say; // from the previous step
 use std::io::{stdout, BufWriter};
 
 fn main() {


yoshuawuyts:
Thanks for asking!

It seems you're running the Rust 2015 edition. In the Rust 2018 edition `extern crate` is no longer required. If you run `cargo new` or `cargo init` for projects, they're set to the 2018 edition by default.

In order to fix this error for an existing project, add:
```toml
[package]
edition = "2018"
```
to your `Cargo.toml` file.

We expect people new to Rust to use the 2018 edition, which means `extern crate` isn't needed.

matt-h:
Thanks, that makes sense.

I had rust 1.29 installed when I started, but when I started going through the getting started and hit a lot of errors I realize I was out of date and upgraded to 1.34 which fixed most of them.

I didn't think to try starting over with a new project and the `Cargo.toml` would be different between versions.

Issue #715, opened 2019-02-24
Rust Security Team public key link is inaccessible


jens1o:
# Content Bug Report

## Page(s) Affected
https://www.rust-lang.org/policies/security

## What needs to be fixed?
The second paragraph links to a public key security researchers are able to encrypt their emails with. 

## Suggested Improvement
We could simply add a text version of the public key given below, or make an anchor to it. (Although personally I'd prefer having a new tab open with the plain key).

yoshuawuyts:
@jens1o could you say more about what's motivating your suggestion?

jens1o:
I wanted to know how I report security issues, finding that the link is no longer accessible?

Sorry, but I'm not really sure where exactly I should elaborate on.

steveklabnik:
The plaintext version of the key is at the bottom of the page, right?

And yeah, that link does 404 so we should fix it.

jens1o:
> The plaintext version of the key is at the bottom of the page, right?

Yeah, there is. I just wanted to say that. I didn't know how this is an suggestion. Just restoring intended behavior.

PR #759, opened 2019-05-13
Add downloadable security team key

skade:
Fixes https://github.com/rust-lang/www.rust-lang.org/issues/715


skade:
This is a tiny change, I'm merging without review.

Issue #746, opened 2019-04-25
Redirect /en-US/ (and other locales) to homepage


badboy:
# Technical Feature Request
## Summary
With the new website, we (unfortunately) dropped translations.
That means that direct links to the index of a locale results in a 404.
This also happens for https://rust-lang.org/en-US/.

At least the last one should redirect to the main page (though we already redirect every `/<locale>/<path>` to `/<path>` if that path was valid on the old site, locale is e.g. `en-US`, `de-DE`). 
PR #597 doesn't seem to handle the `empty path` case.

## Motivation
Lots of links out there linking to https://rust-lang.org/en-US/ (or localized versions)

## Drawbacks
No drawbacks.

PR #763, opened 2019-05-13
Add pdf redirects

### LINKED ISSUE #684: New site is missing pdfs
# Content Bug Report

## Page(s) Affected
https://www.rust-lang.org/pdfs/Rust-Chucklefish-Whitepaper.pdf

## What needs to be fixed?
I don't know how many PDFs were hosted on the old site but there's at least one, https://prev.rust-lang.org/pdfs/Rust-Chucklefish-Whitepaper.pdf, which is a 404 on the new site. I'm also not sure if this was linked from anywhere on the old site (I found the link in a Reddit post).

Either the PDFs should be accessible from the new site, or the new site should redirect to the old for them.

skade:
Fixes #684
Fixes https://github.com/rust-lang/www.rust-lang.org/issues/746

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -215,6 +215,10 @@ fn redirect(dest: redirect::Destination) -> Redirect {
 fn redirect_pdfs(dest: redirect::Destination) -> Redirect {
     Redirect::permanent("/static/pdfs/".to_owned() + dest.uri)
 }
+#[get("/en-US", rank = 6)]
+fn redirect_bare_en_us() -> Redirect {
+    Redirect::permanent("/")
+}
 
 #[get("/<_locale>", rank = 8)]
 fn redirect_bare_locale(_locale: redirect::Locale) -> Redirect {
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -211,6 +211,20 @@ fn redirect(dest: redirect::Destination) -> Redirect {
     Redirect::permanent(dest.uri)
 }
 
+#[get("/pdfs/<dest>")]
```
pietroalbini: Maybe the redirect above could be changed to [`/<dest..>`](https://rocket.rs/v0.4/guide/requests/#multiple-segments) to avoid special-casing PDFs here?

skade: I took that route as the `<dest>` parameter has pretty particular parsing behaviour, currently only dealing with files. I'd like to rewrite the redirect functionality anyways.

skade: https://github.com/rust-lang/www.rust-lang.org/issues/767

skade:
Yeah, I'd like to turn lingering issues into refactors :).

Issue #452, opened 2018-11-29
Jumping / flickering on page load


dbrgn:
# Technical Bug Report

## Your Environment
| Software         | Version(s) |
| ---------------- | ---------- |
| Form Factor      | Desktop |
| Browser Type     | Firefox |
| Browser Version  | 63.0.3 (64-bit) |

## Is this an issue with Rust, CSS, or HTML (or a combination)?

Probably a combination of CSS and webfonts.

## Current and Expected Behavior

When loading or reloading the page, the page elements jump around a lot. It seems that the layout is being processed only after page load. (I initially thought that this was due to some JS based layouting, but I can't seem to find any JavaScript.)

Here's a screen capture of the behavior: https://tmp.dbrgn.ch/rust-2018-11-30_00.04.14.webm

For example, the main menu can quickly be seen in the upper left corner. After a short moment, it "jumps" to the upper right corner.

I think there are multiple stages of page loading (got the screenshots by pressing ESC during page load):

**1. The raw HTML content**

![screenshot](https://tmp.dbrgn.ch/screenshots/20181129235808-cu58wcq3.png)

**2. Some webfonts have been loaded, some CSS rules have been applied**

![screenshot](https://tmp.dbrgn.ch/screenshots/20181129235832-cid5qnyy.png)

**3. Menu moves to upper right corner, but some page elements like the logo haven't loaded yet**

![screenshot](https://tmp.dbrgn.ch/screenshots/20181129235859-gmy0e71o.png)

**4. Page is fully loaded**

![screenshot](https://tmp.dbrgn.ch/screenshots/20181129235925-vzfxpx04.png)

The fact that for example the logo isn't loaded immediately seems to indicate that there are no caching headers set up for static files. Could that be? (see also #319) Would probably good to cache *all* static resources, with proper cache busting parameters appended on deployment.

## Step to Reproduce

Load the page in Firefox, then press F5 or Ctrl+F5.

ashleygwilliams:
this is related to a few open issues currently, but it's an excellent comprehensive issue- so i'm going to opt to close the others in favor of this. thank you for the excellent issue- it's very appreciated.

yoshuawuyts:
Ive seen this behavior before, and I believe the root cause of it is that we don't inline critical CSS inside each page load. To my knowledge Rust currently doesn't have any crates capable of this, so we might need to get a bit creative.

[...]

Issue #539, opened 2018-12-04
name custom css file with sha to unblock caching


ashleygwilliams:
unblocks the small block for #319 re css

PR #768, opened 2019-05-14
Set cache headers on all statics

skade:
Works and sets max-age to 3600 for everything. Blocked on https://github.com/rust-lang/www.rust-lang.org/issues/539

Fixes: #452 #539


skade:
Rebased on current master.

skade:
Merge after https://github.com/rust-lang/www.rust-lang.org/pull/789

PR #773, opened 2019-05-15
Use Tachyons in "get-involved"

### LINKED ISSUE #653: "Get involved" boxes should align buttons at 1024px width
# Design Bug Report

## Page(s) Affected
https://www.rust-lang.org/#read-rust

## What needs to be fixed?
Responsive design

That these buttons (who are black and thus totally highlighted) are not aligned makes it look as if the right block would be "too high", i.e. to much at the top. I get the headlines are aligned, but it still looks like that (on the first look):

image

This only happens when you are around 100px width, in my case 1024px. Note this is a usual size of these smaller, old monitors, so it is not uncommon…

## Suggested Improvement
I'd say you can solve it, if you align the buttons. You can leave some space in the middle, but align the buttons. This is – for most screen widths – also automatically done by the text content, which spans three lines on both sides then:

image

skade:
Fixes #653

### DIFF
--- a/templates/components/panels/get-involved.hbs
+++ b/templates/components/panels/get-involved.hbs
@@ -10,7 +10,7 @@
         <p>We love documentation! Take a look at the books available online, as well as key blog posts and user guides.</p>
         <a href="learn" class="button button-secondary">Read the book</a>
       </div>
-      <div id="watch-rust" class="mw-50-l pt0 flex flex-column justify-between-l">
+      <div id="watch-rust" class="mw-50-l pt3 pt0-l flex flex-column justify-between-l">
         <h3>Watch Rust</h3>
         <p>The Rust community has a dedicated YouTube channel collecting a huge range of presentations and
         tutorials.</p>
--- a/templates/components/panels/get-involved.hbs
+++ b/templates/components/panels/get-involved.hbs
@@ -17,7 +17,7 @@
         <a href="https://www.youtube.com/channel/UCaYhcUwRBNscFNUKTjgPFiA" class="button button-secondary">Watch the Videos</a>
       </div>
     </div>
-    <div>
+    <div class="pt3">
       <h3>Contribute code</h3>
       <p>
       Rust is truly a community effort, and we welcome contribution from hobbyists and production users, from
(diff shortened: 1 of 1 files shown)


templates/components/panels/get-involved.hbs
```diff
@@ -4,20 +4,20 @@
       <h2>Get involved</h2>
       <div class="highlight"></div>
     </header>
-    <div class="row media">
-      <div class="six columns" id="read-rust">
+    <div class="flex flex-column flex-row-l">
+      <div id="read-rust" class="mw-50-l mh2-l pt0 flex flex-column justify-between-l">
```
pietroalbini:
```suggestion
      <div id="read-rust" class="mw-50-l mr4-l pt0 flex flex-column justify-between-l">
```

Using `mh2-l` you also add margin on the left, removing the alignment with the row below:

Istantanea_2019-05-15_19-23-18

If you instead add a `mr4-l` on the left column and no margin on the right one the content will be properly aligned.

templates/components/panels/get-involved.hbs
```diff
@@ -4,20 +4,20 @@
...
         <h3>Read Rust</h3>
         <p>We love documentation! Take a look at the books available online, as well as key blog posts and user guides.</p>
         <a href="learn" class="button button-secondary">Read the book</a>
       </div>
-      <div class="six columns" id="watch-rust">
+      <div id="watch-rust" class="mw-50-l mh2-l pt0 flex flex-column justify-between-l">
```
pietroalbini:
```suggestion
      <div id="watch-rust" class="mw-50-l pt0 flex flex-column justify-between-l">
```

Other part of the misalignment change.

templates/components/panels/get-involved.hbs
```diff
@@ -4,20 +4,20 @@
...
         <h3>Watch Rust</h3>
         <p>The Rust community has a dedicated YouTube channel collecting a huge range of presentations and
         tutorials.</p>
         <a href="https://www.youtube.com/channel/UCaYhcUwRBNscFNUKTjgPFiA" class="button button-secondary">Watch the Videos</a>
       </div>
     </div>
-    <div class="pt3">
+    <div>
```
pietroalbini: In the current website there is a divider between these two rows, but with the PR there is basically no white space between them. Can you add some margin?

The margin should be added also between the other two columns on mobile.

skade: Good catch! Added the spacing back in.

skade:
Right, now I see what you mean. Fixed.

Issue #739, opened 2019-04-09
Set HTTP Strict-Transport-Security header


alex:
# Technical Feature Request
## Summary
Set the `Strict-Transport-Security` header on requests r-l.o, which a long `max-age`.

## Motivation

This ensures that users' browsers automatically take them to the https:// version of the site, even if they type http:// into their browser, thus avoiding an insecure request which can be hijacked.

## Drawbacks
None that I can think of (unless there was a plan to drop https://, but I doubt that :-))

## Rationale and alternatives

This is the web standard way of expressing "this page is always https://". Not doing this leaves a small window of potential content injection.

## Maintenance
It's a single header, maintenance should be limited.

## Unresolved Questions

At the moment it's not possible to use the `includeSubDomains` directive in the header because not all r-l.o subdomains (e.g. ping) support https://. This also means that it's not possible to do HSTS preloading. These concerns can be followed up after an initial HSTS deployment.

pietroalbini:
I'll take on this (along with all the other headers needed to reach an A+ grade on the [Observatory](https://observatory.mozilla.org/analyze/www.rust-lang.org)) as part of an audit of those headers I'm doing across all Rust websites.

Darkspirit:
See also: rust-lang/rustup.rs#180

pietroalbini:
Opened a PR to add all the headers needed to reach A+ on the [Mozilla Observatory](https://observatory.mozilla.org): https://github.com/rust-lang/www.rust-lang.org/pull/778.

> The availability of a ping service can't outweigh a fundamental security feature. It could be still accessed with curl or a second Firefox profile (network.stricttransportsecurity.preloadlist;false).

We won't add the Rust domains on the HSTS preload list right now. I plan to bring the topic up in the infra team once we finish the audit of all the domains.

PR #778, opened 2019-05-18
Add security headers

pietroalbini:
This PR adds all the security headers needed to reach an A+ grade on the [Mozilla Observatory](https://observatory.mozilla.org), including a Content Security Policy. As part of adding the CSP I visited all the pages locally to make sure nothing legit is blocked -- I hope I didn't miss any of the pages.

Fixes https://github.com/rust-lang/www.rust-lang.org/issues/739
r? @skade


skade:
I can't find any additional places with inline styles, so this looks okay to me.

Issue #780, opened 2019-05-20
Frontend cleanup: Remove Skeleton


skade:
Supersedes #183 

The Rust website currently uses two CSS frameworks: Skeleton and Tachyons (http://tachyons.io/docs/layout/flexbox/). This is due to an unfinished migration. This unfinished migration is a blocker for further extensions of the website.

We'd like to completely move over to Tachyons. There's two wishes here:

* We should avoid tachyons _grid_ and instead us its flexbox classes: http://tachyons.io/docs/layout/flexbox/
* If possible, elements of similar shape should use the same code.

It's easy to help out! Here are some instructions:

* Check out the website
* Pick _one_ of the subpages
* Post here that you care for that subpage. You will have 7 days to post a PR (otherwise, we'll ping you)
* (optional) join https://discord.gg/rust-lang and go to #website if you need help or feedback.
* remove skeletons from the page: https://github.com/rust-lang/www.rust-lang.org/pull/774/files
* Fix what's broken, using Tachyons or custom CSS. Also try to thoroughly remove now unused CSS classes.
* Optional: check this bugtracker for issues concerning you page and try to fix them on the go

Handy page list for everyone to check:

| URL                         | Page Name       | Assigned   | Fixed
| ---                         | ---             | ---        | ---
| `/`                         | Index           |  @dzhibas         | yes
| `/install`                | Install           | @andirotter | yes
| `/learn`                    | Learn           | @Pjb518             |  yes
| `/tools`                    | Tools           | @dzhibas           |  yes
| `/governance`               | Governance      |   @dzhibas         | yes
| `/community`                | Community       | @yayoc           |  yes
| `/what`                     | What            |  @skade          |  yes
| `/what/cli`                 | CLI             |  @skade      |  
| `/what/embedded`            | Embedded        |  @skade           |  
| `/what/wasm`                | WASM            |   @skade          |  
| `/what/networking`          | Server          |  @dzhibas           |  yes
| `/what/production`          | Production      |   @dzhibas       | yes
| `/policies`                 | Policies        |     @chriskrycho       | yes
| `/policies/media-guide`     | Media Guide     |   @chriskrycho         | yes
| `/policies/security`        | Security        |   @chriskrycho         | yes
| `/policies/license`         | License         |     @chriskrycho      |  yes
| `/policies/code-of-conduct` | Code of Conduct |  @chriskrycho          |  yes

andirotter:
Hey there, 

[...]

PR #787, opened 2019-05-21
#780 get rid of grid, moving to tachyons flex for /tools

dzhibas:
checked multiple breakpoints for /tools, now looks good


skade:
Thanks, this indeed looks good!

PR #544, opened 2018-12-04
[WIP] feat(styles): name compiled sass with sha

ashleygwilliams:
#️⃣ do not merge #️⃣ 

fixes #539

PR #789, opened 2019-05-21
Implement hashing for CSS filenames

### LINKED ISSUE #539: name custom css file with sha to unblock caching

skade:
Based on @ashleygwilliams  work in #544

Uses lazy_static as threading this through as _state_ is painful, as catchers and routes have a different interface here.

Fixes #539

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,10 @@
 /target/
 **/*.rs.bk
 /node_modules
-/static/styles/app.css
-/static/styles/fonts.css
 /static/styles/*.map
 .sass-cache
 localhost*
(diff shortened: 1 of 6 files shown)


skade:
Rebased on current master.

skade:
@pietroalbini implemented and rebased.

I called the js file `app_<hash>.js`, even though it's currently only used on the install page. If we need more JS at other places in the page, it probably still makes sense to ship it as one in one go instead of per page, I found the name more fitting there.

pietroalbini:
Thanks!

PR #790, opened 2019-05-21
Small fixes to #732

skade:
We seem to run into an upgrade problem and I accidentally made a mistake with incorrect formatting.


ashleygwilliams:
reverting due to ops incident.

PR #794, opened 2019-05-22
Move homepage and common elements into translation file

Manishearth:
Also fills in xx-AU testing locale

PR #796, opened 2019-05-23
Governance internationalization

Manishearth:
Based on https://github.com/rust-lang/www.rust-lang.org/pull/794

Needs https://github.com/rust-lang/team/pull/70

This is a tad more involved since the team data comes from https://github.com/rust-lang/team


Manishearth:
Ready for review. It's fine to just review this PR all at once and land, but if you want to do it piecewise it may make sense to land #794 first

Manishearth:
I _think_ with this PR the rest of the i18n is completely mechanical -- going through the templates, adding localization strings, and moving on. Once this lands we can also start asking folks to translate the ftl files (ideally pontoon is set up, but it's not necessary)

Issue #784, opened 2019-05-20
Don't hide text during webfont load


polarathene:
# Technical Bug Report
## Your Environment
| Software         | Version(s) |
| ---------------- | ---------- |
| Form Factor      | Desktop |
| Browser Type     | Chrome - Linux |
| Browser Version  | 73 |

## Is this an issue with Rust, CSS, or HTML (or a combination)?
CSS / HTML - Font loading
## Expected Behavior
Text should not be hidden due to webfont loading.

## Current Behavior
As shown [here](https://www.webpagetest.org/video/compare.php?tests=190520_8C_c03a8eb9be1c4f29c3f88e3b90a204fe-r:1-c:0), the text content is not visible until the staggered loads of webfonts complete.

## Step to Reproduce
Run the same webpagetest linked above(policies URL on a slow connection)

## Screenshots
![screenshot](https://www.webpagetest.org/results/19/05/20/8C/c03a8eb9be1c4f29c3f88e3b90a204fe/video_1/ms_005466.jpg)

skade:
@polarathene could you please check if #797 improved the situation for you? (allow ~10 minutes for deployment)

If yes, could you please close this issue?

polarathene:
@skade Definitely better with text being there straight away. There is a little bit of reflow going on due to the differences between fonts swapped, mostly letter-spacing/kerning I think? Most noticeable difference is the `h1`s which have a big contrast in font-weight, I guess there isn't much you can do about that.

PR #797, opened 2019-05-23
Change font-display to swap

BO41:
Changing the [`font-display`](https://css-tricks.com/almanac/properties/f/font-display/#article-header-id-0) property to `swap` instead of `block` improves the time until text is rendered.

Helps (or even closes) #784


skade:
Thanks, let's try that out!

Issue #798, opened 2019-05-23
[meta] Add localization strings everywhere


Manishearth:
Part of https://github.com/rust-lang/www.rust-lang.org/issues/799

With https://github.com/rust-lang/www.rust-lang.org/pull/794 and https://github.com/rust-lang/www.rust-lang.org/pull/796 we have all of the infrastructure necessary to add localization strings everywhere.

If you're interested in helping, check out the i18n branch, pick an unclaimed file from the below list, and follow the instructions below.

Files needing creation (and all the template files they should contain info for):

 - [x] `learn.ftl`: https://github.com/rust-lang/www.rust-lang.org/tree/i18n/templates/learn (@Manishearth , #816)
 - [x] `cli.ftl`: https://github.com/rust-lang/www.rust-lang.org/blob/i18n/templates/what/cli.hbs, https://github.com/rust-lang/www.rust-lang.org/tree/i18n/templates/components/what/cli (@eignnx, #803)
 - [x] `embedded.ftl`: https://github.com/rust-lang/www.rust-lang.org/blob/i18n/templates/what/embedded.hbs, https://github.com/rust-lang/www.rust-lang.org/tree/i18n/templates/components/what/embedded (@eignnx , #822)
 - [x] `networking.ftl`: https://github.com/rust-lang/www.rust-lang.org/blob/i18n/templates/what/networking.hbs, https://github.com/rust-lang/www.rust-lang.org/tree/i18n/templates/components/what/networking
 - [x] `wasm.ftl`: https://github.com/rust-lang/www.rust-lang.org/blob/i18n/templates/what/wasm.hbs, https://github.com/rust-lang/www.rust-lang.org/tree/i18n/templates/components/what/wasm (@eignnx, #824 )
 - [x] `policies.ftl`: https://github.com/rust-lang/www.rust-lang.org/tree/i18n/templates/policies (@eignnx, #813)

 - [x] `tools.ftl`: https://github.com/rust-lang/www.rust-lang.org/tree/i18n/templates/tools (@eignnx, #800, #801)

Leave a comment if you're interested in picking up one of these!

Manishearth:
## Instructions

Make sure you're working off the `i18n` branch.

The way this works is that you take one of the handlebars files under `templates/`, and move all of their strings into a Fluent .ftl file. So, you start off with something like

```hbs
# templates/pitch.hbs
<div>
<h2>The Rust programming language is super cool</h2>
[...]

PR #813, opened 2019-05-25
Move Text from Policy Pages into Translation Files

eignnx:
Work done for #798 

## Change List:
User-facing text from the following files was extracted into translation files:
* `templates/policies/index.hbs` extracted into `policies.ftl`
* `templates/policies/code-of-conduct.hbs` extracted into `code-of-conduct.ftl`
* `templates/policies/licenses.hbs` extracted into `licenses.ftl`
* `templates/policies/media-guide.hbs` extracted into `media-guide.ftl`
* `templates/policies/security.hbs` extracted into `security.ftl`

## Changes in my Process
[REMOVED, see [reasoning](https://github.com/rust-lang/www.rust-lang.org/pull/813#issuecomment-496017240)]

## Testing
I have visually inspected all the pages modified, comparing the modified versions to the corresponding live https://www.rust-lang.org pages. I have not spotted any differences, but maybe some sort of HTML diffing tool could be used to verify that nothing has changed?

### DIFF
--- a/locales/en-US/code-of-conduct.ftl
+++ b/locales/en-US/code-of-conduct.ftl
@@ -16,8 +16,8 @@ coc-conduct-description =
           <li>Private harassment is also unacceptable. No matter who you are, if you feel you have been or are being harassed or made uncomfortable by a community member, please contact one of the channel ops or any of the { $coc-rust-moderation-team-anchor } immediately. Whether you’re a regular contributor or a newcomer, we care about making this community a safe place for you and we’ve got your back.</li>
           <li>Likewise any spamming, trolling, flaming, baiting or other attention-stealing behavior is not welcome.</li>
         </ul>
-coc-rust-moderation-team-anchor-text = Rust moderation team
-coc-conduct-link = Email The Moderation Team
+coc-conduct-description-team-anchor-text = Rust moderation team
+coc-conduct-email-button = Email The Moderation Team
 
 coc-moderation-heading = Moderation
 coc-moderation-description =
@@ -36,3 +36,5 @@ coc-moderation-description =
         <p>And if someone takes issue with something you said or did, resist the urge to be defensive. Just stop doing what it was they complained about and apologize. Even if you feel you were misinterpreted or unfairly accused, chances are good there was something you could’ve communicated better — remember that it’s your responsibility to make your fellow Rustaceans comfortable. Everyone wants to get along and we are all here first and foremost because we want to talk about cool technology. You will find that people will be eager to assume good intent and forgive as long as you earn their trust.</p>
         <p>The enforcement policies listed above apply to all official Rust venues; including Discord channels (https://discord.gg/rust-lang); GitHub repositories under rust-lang, rust-lang-nursery, and rust-lang-deprecated; and all forums under rust-lang.org (users.rust-lang.org, internals.rust-lang.org). For other projects adopting the Rust Code of Conduct, please contact the maintainers of those projects for enforcement. If you wish to use this code of conduct for your own project, consider explicitly mentioning your moderation policy or making a copy with your own moderation policy so as to avoid confusion.</p>
         <p><i>Adapted from the <a href="http://blog.izs.me/post/30036893703/policy-on-trolling">Node.js Policy on Trolling</a> as well as the <a href="https://www.contributor-covenant.org/version/1/3/0/">Contributor Covenant v1.3.0</a>.</i></p>
+
+coc-moderation-description-team-anchor-text = Rust moderation team
--- a/locales/en-US/code-of-conduct.ftl
+++ b/locales/en-US/code-of-conduct.ftl
@@ -1,0 +1,38 @@
+### Translation file for page: https://www.rust-lang.org/policies/code-of-conduct
+
+## templates/policies/code-of-conduct.hbs
+
+coc-page-heading = Code of conduct
+
+coc-conduct-heading = Conduct
+coc-conduct-description =
+        <ul>
+          <li>We are committed to providing a friendly, safe and welcoming environment for all, regardless of level of experience, gender identity and expression, sexual orientation, disability, personal appearance, body size, race, ethnicity, age, religion, nationality, or other similar characteristic.</li>
+          <li>Please avoid using overtly sexual aliases or other nicknames that might detract from a friendly, safe and welcoming environment for all.</li>
+          <li>Please be kind and courteous. There’s no need to be mean or rude.</li>
+          <li>Respect that people have differences of opinion and that every design or implementation choice carries a trade-off and numerous costs. There is seldom a right answer.</li>
+          <li>Please keep unstructured critique to a minimum. If you have solid ideas you want to experiment with, make a fork and see how it works.</li>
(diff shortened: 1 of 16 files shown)


templates/fluent-resource/en-US/security.ftl
```diff
@@ -0,0 +1,112 @@
...
+security-reporting-li3 =
+        Post on the <a href="https://internals.rust-lang.org/">internals forums</a>
+security-reporting-p5 =
+        Please note that the discussion forums are public areas. When escalating
+        in these venues, please do not discuss your issue. Simply say that
+        you’re trying to get a hold of someone from the security team.
+
+## Security coordinator email addreses and links to their public keys
+security-public-key = public key
+-security-coordinator-email-anchor =
```
eignnx: These kinds of links could potentially be pulled out into [`{ $fluent-variables }`](https://projectfluent.org/fluent/guide/variables.html) in the future?

templates/fluent-resource/en-US/security.ftl
```diff
@@ -0,0 +1,112 @@
...
+
+security-reporting-heading = Reporting
+security-reporting-link = email security@rust-lang.org
+security-reporting-p1 =
+        Safety is one of the core principles of Rust, and to that end, we would
+        like to ensure that Rust has a secure implementation. Thank you for
+        taking the time to responsibly disclose any issues you find.
+security-reporting-p2 =
+        All security bugs in the Rust distribution should be reported by email
+        to <a href="mailto:security@rust-lang.org">security@rust-lang.org</a>.
```
skade: I would avoid having the email and the links for the security team to be changeable by translation. I see a slight vector there where a translation could be modified to say something else.

I'm not sure if I'm just paranoid there.

Same does, btw. apply to policy pages of all kind where core or team approval is necessary.

eignnx: I can definitely extract these links into [referencing messages](https://projectfluent.org/fluent/guide/references.html) for now. I'm not sure how to set them up as [variables](https://projectfluent.org/fluent/guide/variables.html) that get injected by the server, but that seems like a better way to do it in the long run.

> . . . a translation could be modified to say something else.

Isn't this a concern with all translation work though? I agree that variables at least gives us some assurances that typos weren't made by accident.

Manishearth: @eignnx stick them in core.ftl for now. I'll move them out into a common file later.

templates/fluent-resource/en-US/security.ftl
Manishearth: Yeah, please move these all into `core.ftl`.

I'll rejigger this and move them outside of the language folders later.

templates/fluent-resource/en-US/licenses.ftl
```diff
@@ -0,0 +1,36 @@
...
+        any of its associated projects are licensed, please contact the Rust
+        Core Team.
+licenses-license-email-link = Email The Core Team
+
+licenses-attribution-heading = Attribution
+licenses-attribution-description =
+        Icons made by { licenses-freepik-anchor } from { licenses-flaticon-anchor },
+        licensed by { licenses-cc-by-3-0-anchor }
+
+licenses-freepik-anchor =
```
Manishearth: I think these can just be put inline, no need for variables

templates/fluent-resource/en-US/code-of-conduct.ftl
```diff
@@ -0,0 +1,113 @@
+### Translation file for page: https://www.rust-lang.org/policies/code-of-conduct
+
+## templates/policies/code-of-conduct.hbs
+
+coc-page-heading = Code of conduct
+
+coc-conduct-heading = Conduct
+coc-conduct-li1 =
```
Manishearth: The entire HTML code of conduct should be a single key

templates/fluent-resource/en-US/code-of-conduct.ftl
```diff
@@ -0,0 +1,113 @@
...
+coc-conduct-li8 =
+        Likewise any spamming, trolling, flaming, baiting or other
+        attention-stealing behavior is not welcome.
+coc-conduct-link = Email The Moderation Team
+
+coc-moderation-heading = Moderation
+coc-moderation-p1 =
+        These are the policies for upholding our community’s standards of
+        conduct. If you feel that a thread needs moderation, please contact the
+        <a href="{{baseurl}}/governance/teams/moderation">Rust moderation
```
Manishearth: Handlebars templates won't work here, make this a `{ $fluent-variable }` and use `textparam` to pass it in.

Manishearth:
> In previous i18n commits I have been extracting entire <p> and <ul> tags into the FTL files (see these two comments), but this time I decided to try leaving the <p> and <ul> tags and just extracting the text. This means there are many more Fluent messages (i.e. variables) in some files, but each message is smaller in size. I made this process change because I think we should expose translators to as little raw HTML as we can, and it's not too much more work to split adjacent paragraphs into multiple variables.

So I like the sentiment here -- indeed I was considering doing this too when I started; and this is why [this segment](https://github.com/rust-lang/www.rust-lang.org/blob/83c6453d3e3ce282ac493a8df84f88506f74c8b6/templates/fluent-resource/en-US/tools.ftl#L1) uses fluent variables.

But when talking with folks working on Fluent it seems like translators are typically expected to use a _little_ HTML (Fluent.js has a DOM overlay feature for making this easier, but Fluent.rs doesn't yet). So `<p>` and `<li>` should be fine.

templates/fluent-resource/en-US/security.ftl
```diff
@@ -0,0 +1,112 @@
...
+        (alternatively by sending an email to
+        { -rustlang-security-announcements-email-anchor }). The mailing list is
+        very low traffic, and it receives the public notifications the moment
+        the embargo is lifted.
+security-receiving-p2 = 
+        We will announce vulnerabilities 72 hours before the embargo is lifted
+        to { -distros-openwall-email-anchor }, so that Linux distributions can
+        update their packages.
+
+## Security mailing list links and emails
```
Manishearth: Move these to core.ftl.

(I'm going to move the `rust` key into `common.ftl` and then move `core.ftl` outside of the locale folders)

eignnx: Can do. The text "Rust security announcements mailing list" needs to be exposed to translators, so I'm thinking I'll leave the `security-rust-security-announcements-mailing-list` variable (line 103) in `security.ftl`.

Manishearth: Sounds good, the email address should be in core

templates/fluent-resource/en-US/code-of-conduct.ftl
eignnx: @Manishearth Just checking, but I'm assuming this doesn't apply to headings? There are two sections to the COC, and `h3` tags have ids on them. Does the whole page need to be stored as a unit for legal reasons?

Manishearth: Depends on the case, but I would treat the moderation policy as a separate pair of keys (header + text)

Overall we should be very careful about complex formatting making its way in (like, class names, etc) but simple tags are fine.

This is a bit of an exercise in balance, in the case of a bajillion bullet items the whole thing should be a single key, but if there are just two it's fine to split.

eignnx:
@Manishearth ready for review, but I think I need help rebasing

Manishearth:
Done

Manishearth:
I'm also going to squash the commits down, and fix the core.ftl stuff (soon)

Manishearth:
@eignnx you didn't need to make link refs for things like internals.rust-lang.org, fwiw. But if they exist that's fine too

Manishearth:
There was a single ref that needed updating in the coc file, fixed that. Everything else is identical. Thanks so much for doing this!

PR #815, opened 2019-05-26
Add i18n fallback

renyuneyun:
A better handler for i18n fallback, using fluent-locale.

This shall separate the logic for fallback, and make more advanced fallback implemented easier.


renyuneyun:
One personal expectation is to unify most (if not all) Chinese resources, taken the advantage of an automatic converter [OpenCC](https://github.com/BYVoid/OpenCC) ([rust binding](https://github.com/magiclen/opencc-rust)).

OpenCC supports the convert between Simplified and Traditional characters as well as regional term (Mainland, Hong Kong, Taiwan, etc). It generally works well (it's the best one I ever see).
But there are a few cases it may not work well (due to the nature of this task (I can explain in more details if needed)). Localized resources can then be added (with higher priority) when automatic convert doesn't work well (after people reported).

Manishearth:
Can you implement this on top of the `i18n` branch? A lot has changed. I plan to merge that into master at some point, but for now we're doing all the internationalization work there so that we don't accidentally break the main website.

Also, we'll have to have a hardcoded list of supported languages (for the dropdown) anyway, so we should probably structure this around that.

renyuneyun:
@Manishearth Ok, no problem.
I'll rebase it now, and then try to use reference.

renyuneyun:
@Manishearth I see in the code (both master and i18n branch) that the `RESOURCES` and `BUNDLES` are created using the older way (and the lifetime for a lot of things is explicitly specified as always `'static`). Is this due to the bug your previously mentioned?

Would that be a good idea or a bad idea to minimize that piece of lifetime in struct's and function's definitions?

Manishearth:
We don't want to minimize it. It's a static object. Being `'static` is deliberate, it's much easier to work with.

We may want to make it possible to reload files while running, but that's a significant amount of work, and it has little benefit given our current deployment model.

Manishearth:
Thanks!

Manishearth:
This needs to be tweaked to deal with the governance stuff correctly, though.

Manishearth:
Oh, I realized that this approach won't work for the governance at all: We shouldn't use the fallback chain for defaulting to English, because our English default works differently in different places.

Furthermore, we should definitely not be recomputing this each time. I've got a plan.

Manishearth:
Also, we shouldn't be using lifetimes here, again :smile:

PR #818, opened 2019-05-26
Add learn/get-started

Manishearth:


locales/en-US/learn.ftl
```diff
@@ -61,3 +61,77 @@ learn-nomicon-button = Read the ’nomicon
 learn-unstable-alt = A hand sharing sparkles
 learn-unstable = The Unstable Book has documentation for unstable features that you can only use with nightly Rust.
 learn-unstable-button = Read the unstable book
+
+# learn/get-started.hbs
```
eignnx: These kinds of comments are seen by translators. I would suggest using two pound signs for FTL file section labels. See https://projectfluent.org/fluent/guide/comments.html

locales/en-US/learn.ftl
```diff
@@ -61,3 +61,77 @@ learn-nomicon-button = Read the ’nomicon
...
+    <p><code>src/main.rs</code> is where we’ll write our application code.</p>
+    <hr/>
+    <p><code>cargo new</code> generates a "Hello, world!" project for us! We can run this program by moving into the new directory that we made and running this in our terminal:</p>
+    <p><code>cargo run</code></p>
+    <p>You should see this in your terminal:</p>
+    { $output }
+
+learn-dependencies = Adding dependencies
+learn-dependencies-steps = <p>Let’s add a dependency to our application. You can find all sorts of libraries on <a href="https://crates.io">crates.io</a>, the package registry for Rust. In Rust, we often refer to
+      packages as “crates.”</p>
```
eignnx: Could this linebreak be removed? Random linebreaks like this look kinda weird in Pontoon especially since the line above is so long.

Manishearth:
Fixed in a commit after the fact

PR #819, opened 2019-05-26
Parameterize {{baseurl}} in install.hbs file

eignnx:
Work for #798 

I `grep`'d for `{{baseurl}}` inside of FTL files, and this was the last instance. It has now been extracted into a `textparam`.


Manishearth:
Also uplifted to the master branch

PR #820, opened 2019-05-26
Move core strings outside of en-US

Manishearth:
These should never be edited by localizers

cc @eignnx

r? @skade


eignnx:
Should be fine, nothing really to translate in core.ftl, and it doesn't reference any external messages.

Manishearth:
Pontoon may not like it, but we'll see!

PR #822, opened 2019-05-27
Move Text from Embedded page into Translation Files

eignnx:
Work for #798 

The webpage has been visually inspected side-by-side with the live site, and no differences have been found. I compared:
* Link `href`s
* Paragraph wrap and spacing
* Image alt text


eignnx:
@Manishearth ready for review

PR #828, opened 2019-05-28
Fix localization error on learn/Getting Started

eignnx:
One line fix for this problem happening on https://www.rust-lang.org/learn/get-started:

edit: OOPS FIXED SUBTITLE:
image


eignnx:
@Manishearth r?

Issue #758, opened 2019-05-13
Bad contrast on code examples


skade:
# Design Bug Report

## Page(s) Affected
All pages with code:
* [ ] https://www.rust-lang.org/policies/security
* [ ] https://www.rust-lang.org/learn/get-started
* [ ] https://www.rust-lang.org/tools
* [ ] https://www.rust-lang.org/learn/get-started
* [ ] https://www.rust-lang.org/tools/install

## What needs to be fixed?
The code examples need to get a better visibility. Often, they are white on transparent white with a background color. The code highlighting works great for white background.

## Suggested Improvement
Find a better display method for the code examples.

![Screenshot from 2019-05-15 10-51-56](https://user-images.githubusercontent.com/47542/57762368-aea07480-76ff-11e9-8299-bb01aae7bc6b.png)
![Screenshot from 2019-05-15 10-51-14](https://user-images.githubusercontent.com/47542/57762369-aea07480-76ff-11e9-834d-7769cd90cd2f.png)
![Screenshot from 2019-05-15 10-51-02](https://user-images.githubusercontent.com/47542/57762370-aea07480-76ff-11e9-8909-a97fbdc15a32.png)
![Screenshot from 2019-05-15 10-50-11](https://user-images.githubusercontent.com/47542/57762371-af390b00-76ff-11e9-8d53-8a2d179548f8.png)
![Screenshot from 2019-05-15 10-49-17](https://user-images.githubusercontent.com/47542/57762372-af390b00-76ff-11e9-9553-954492b6016d.png)

skade:
Portential PR here: https://github.com/rust-lang/www.rust-lang.org/pull/606

skade:
PR #606 has the problem of making all background forcibly white, which puts "holes" into the design. We basically need to create out own themes there.

As an initial fix, we'd prefer a non-highlighted change, which tunes the backgrounds.

skade:
This is fixed by #831, #764 is the tracking issue for syntax highlighting.

PR #831, opened 2019-05-29
fix for Issue #758 Bad contrast on code examples

madmalik:
This change does two things:
- The code blocks are set on a darker background. I've tested a unified color scheme for all color-contexts, but for me it pushed the already noisy site over the edge. There is still no syntax highlighting, but with the darkened background there is enough headroom in terms of contrast to use the link-yellow and a lighter variant of another color in the color schema.
- I unified the paddings of the code blocks so that all inline and full width elements get the same values respectively and i did a small refactoring of the mixins that where used for that.

This is the smallest change to make the code blocks usable without deviating too much from the old design. Imo, larger changes/redesigns of the code blocks should be done in context of the overall design.

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -62,27 +62,18 @@ pre > code {
   white-space: pre;
 }
 
-@mixin code-space-full-width {
-  padding: 1rem 1.5rem;
-  margin: 0 .2rem;
-}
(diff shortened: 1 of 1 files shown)


skade:
LGTM, thanks!

PR #835, opened 2019-05-29
Escape HTML characters in code examples

skade:
Fixes https://github.com/rust-lang/www.rust-lang.org/issues/429 the hard way.

I think that's feasible for now, it's better then leaving the issue open.


skade:
I agree.

skade:
I agree.

nikhiljha:
Full circle to #465 

:)

Issue #851, opened 2019-06-02
Image title contains a `<span>`


kennytm:
# Design Bug Report

## Page(s) Affected

The front page of www-staging (both http://www-staging.rust-lang.org/ and http://www-staging.rust-lang.org/tr)

## What needs to be fixed?

The "title" text of the Twitter, Discord and GitHub icon under Social section contains a literal `<span>` tag.

<img width="373" alt="Screenshot 2019-06-02 at 22 43 56-fs8" src="https://user-images.githubusercontent.com/103023/58763031-b4fc7080-8588-11e9-9425-79bc40321c0f.png">

## Suggested Improvement

Remove the `<span>` from the title.

PR #852, opened 2019-06-02
Don't attach incontext tags for alt tags

Manishearth:
Fixes https://github.com/rust-lang/www.rust-lang.org/issues/851

r? @skade @kennytm


Manishearth:
Also merging this to trigger a change.

PR #853, opened 2019-06-02
Add lang attribte

Manishearth:
This fixes things like capitalization in Turkish, and will also prevent Han unification issues.


Manishearth:
Going to merge this early so there's something to git sync

PR #855, opened 2019-06-04
Add missing context in production page

Danue1:


locales/en-US/production.ftl
```diff
@@ -30,5 +31,3 @@ production-contact-core = Message the Core Team
 production-license-header = License
 production-license = Rust is dually licensed under the Apache License Version 2.0 and the MIT license, making it suitable for all sorts of projects.
 production-license-link = Policies: License
-
-production-users = Users
```
Manishearth: Please don't remove this entry, it's used in production/users.hbs

PR #856, opened 2019-06-04
Support language-switching

Manishearth:
r? @skade


Manishearth:
merging early so we can play with this on staging

PR #857, opened 2019-06-05
Use `lang` attributes where appropriate

Manishearth:
Added some helpers that force things to be lang=en-US for correct rendering

r? @skade


Manishearth:
Going to merge immediately so that the French translators see these changes.

PR #859, opened 2019-06-05
Localize page titles

Manishearth:
r? @skade


Manishearth:
Early merging again, but please review later before deploy.

PR #865, opened 2019-06-06
Fix learn-doc-std-button text

timClicks:
Changes the label from "The standard" to "The standard library" to prevent visitors to the site thinking that they'll be taken to a spec of the language

Screenshot of [current page](https://rust-lang.org/learn):

rust-learn


Manishearth:
Yipes.

Manishearth:
@vertexclique @sanxiyn @lifthrasiir would y'all be able to fix the Turkish and korean translations for this? (in Pontoon)

PR #871, opened 2019-06-10
Alt text of `chip2.svg` doesn't match actual image

shamiao:
1. `TFQP` is a typo, should be `TQFP`. Also, the image has 20 pins instead of 16. 
2. `TQFP` (Thin-QFP) is a specific package, but this image is abstract, we simply cannot tell which specific QFP this image is. 

The correct alt text of <https://www.rust-lang.org/static/images/chip2.svg> is "QFP-20 package".


Manishearth:
Thanks, I'll also fix the translations.

Issue #830, opened 2019-05-29
Images on WASM page are a little too small


skade:
# Content Bug Report

## Page(s) Affected
https://www.rust-lang.org/what/wasm#get_started

## What needs to be fixed?
The images and items are correctly aligned, but the images are a little too small on desktops.

## Suggested Improvement
Make images a little bigger.

skade:
Fixed by #874

PR #874, opened 2019-06-10
resizes and aligns wasm images (#830)

JonathanMH:
Hi!

This is a pull request potentially fixing both the alignment and the sizing (#830) issues on the wasm page.

The added classes controlling height seemed appropriate to not have to version images in different crops and aspect ratios.

This is my first time using tachyons, feedback very welcome!

rust-image-align-fix


skade:
That works, thanks!

PR #885, opened 2019-06-23
Build on Windows with VS2019

bltavares:
This commit contains a temporary fork on `sass-rs` which provides the
necessary chagnes to have the project running on Windows 10.

I expect [the change](https://github.com/compass-rs/sass-rs/pull/49)
to be taken upstream, so we can simply upgrade the dependency instead of
keeping the forked git version, but this is more so to let others
compile the project and document the error, if someone tries to run it
on Windows meanwhile.

The error message in itself, to help with searchs, is:

```
error MSB4019: The imported project "C:\Microsoft.Cpp.Default.props" was
not found. Confirm that the path in the <Import> declaration is correct,
and that the file exists on disk.
```

This also provides a documentation on the README on what to do in case
the error happens.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,13 +16,17 @@ rocket = "0.4.1"
 serde = "1.0"
 serde_derive = "1.0"
 serde_yaml = "0.8.7"
-sass-rs = "0.2.1"
+# sass-rs = "0.2.1"
 reqwest = "0.9.5"
 toml = "0.4"
 serde_json = "1.0"
 rust_team_data = { git = "https://github.com/rust-lang/team" }
 handlebars = "1.1.0"
 siphasher = "0.2.3"
+
+[dependencies.sass-rs]
+git = "https://github.com/bltavares/sass-rs"
+branch = "compile-on-windows-vs2019"
 
 [dependencies.rocket_contrib]
 version = "*"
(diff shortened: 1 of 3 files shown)


skade:
Hm, this brakes on Travis for linux, though...

README.md
```diff
@@ -12,6 +12,13 @@
 1. To build the app and run the server, run `cargo watch -x run` in your terminal.
 1. Navigate to http://localhost:7878 in your browser. If you make any updates to the styles, or any other project files, `cargo watch` will automatically restart the server for you, all you have to do is refresh your browser to see your changes.
 
+#### Building on Windows
+
+If you have issues comiling `sass-sys` on Windows 10, try the following:
```
skade: `comiling` -> `compiling`

Cargo.toml
```diff
@@ -16,14 +16,18 @@ rocket = "0.4.1"
 serde = "1.0"
 serde_derive = "1.0"
 serde_yaml = "0.8.7"
-sass-rs = "0.2.1"
+# sass-rs = "0.2.1"
```
skade: As far as I see,`sass-rs` has been updated, so we can actually use release.

skade:
Thanks for the patch and the effort, I left you a review!

XAMPPRocky:
Since this change has since been merged upstream to `sass-rs`, I'm going to close this.

PR #895, opened 2019-07-03
Add ckb to the locales directory

realRojSerbest:
# Hi!
### I need your help! I want you to duplicate en-US directory and rename it to ckb I will edit the files in rojserbest.github.io/www.rust-lang.org to translate it to CKB (Central Kurdish Branch).


Manishearth:
We have a process for translations, please see https://github.com/rust-lang/www.rust-lang.org/blob/master/TRANSLATIONS.md . You will have to form a team and use pontoon

PR #911, opened 2019-07-26
add Geany to list of editors that support Rust

andy5995:


skade:
Hey, thanks for the PR, let's add it!

PR #921, opened 2019-09-04
Remove unavailable websites from Rust users

simeg:
These were verified with a script I wrote available at
  https://github.com/simeg/rust-www-friends-verifier


skade:
Thanks!

PR #922, opened 2019-09-14
Improve section title legibility 

XAMPPRocky:
This PR changes the font for `h2`'s to use Fira Sans Extra Bold, over the Alfa Slab One typeface. The reasoning for this is that I find Alfa Slab One to be hard to read outside of large title type. The spacing is incredibly tight, to the point that it can be hard to read even on high resolution displays. I initially attempted to only change the spacing, however Alfa Slab One looks best when it is compressed, when spaced out the letters feel far from each other.

So I tried changing the typeface to Fira Sans, and I found it to be a significant improvement. In order to maintain a visual hierarchy I also reduced the weight of `h3` and `h4`'s to Semi Bold, so that they were still clearly distinct from both the body text and their headers.

I've included screenshots below of some of the main areas this change impacts. 

### Landing page
homepage-diff

### CLI feature list
font-change-list

### Footer
footer-diff

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -115,7 +115,6 @@ code.code-header {
 }
 
 header h1, section h2 {
-  font-family: $header-font;
   z-index: 999999;
   position: relative;
   letter-spacing: 1px;
@@ -123,11 +122,13 @@ header h1, section h2 {
 }
 
 header h1 {
+  font-family: $header-font;
   font-size: 8rem;
   margin-bottom: 0;
   margin-top: 0;
   line-height: 1.2;
   font-weight: 300;
+  letter-spacing: 1px;
 }
 
 header h2 {
@@ -184,7 +185,13 @@ section {
     }
   }
 
+  h2 {
+      font-family: $body-font;
+      font-weight: 800;
+  }
+
   h3 {
+    font-weight: 600;
     line-height: 1.3;
   }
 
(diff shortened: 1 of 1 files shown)


Manishearth:
cc @skade iirc our font choice was influenced by language portability issues?

XAMPPRocky:
@Manishearth Could you expand on what you mean by language portability? This PR does not add or remove any fonts or typefaces.

Manishearth:
Some of the fonts don't work well for some languages for some contexts. Section headers _might_ be one of those, I don't remember.

skade:
We use a patched Fira with extended character sets at most places now. Alfa Slab is kinda rusts "house font", but has huge problems with alternative character sets.

I'm kind of hesitant to moving everything to one font, though, I'm a little conflicted there.

XAMPPRocky:
@skade This PR still keeps Alfa Slab One as the typeface for all titles/`h1`s. It's only sub heading that are affected.

skade:
@XAMPPRocky ugh, sorry, I didn't check that proper. TBQH, the h1 have a similar problem.

XAMPPRocky:
@skade I agree. Alfa Slab has the problem of not having only a single typeface, so it is difficult to use across a website, if there were more varying weights it might be not be an issue. It is my belief that in the long term we'd want to consider choosing a font with a greater variety and extended character set. The process of doing that is something that would take quite awhile to reach consensus on. This PR is intended to be a small quality of life change to improve the current situation until a much more comprehensive solution can be found.

skade:
@xampprocky Okay, I'll check the PR later and will probably merge it.

PR #925, opened 2019-09-22
Link updated for Eclipse Rust (Sep/2019)

amargarido:
Just updating the link. Otherwise it would take the user to the full list of downloads available on Eclipse.org.


skade:
Thanks!

PR #927, opened 2019-10-03
Fix text overflow problem in code examples

### LINKED ISSUE #926: Code example overflow on Networking page
# Design Bug Report

## Page(s) Affected
[Networking](https://www.rust-lang.org/what/networking)

## What needs to be fixed?
Code examples in Get Started section of the Networking page

## Suggested Improvement
Make it so that the code wraps instead of overflowing the page.

Desktop:
image
Mobile:
image

EDIT: After further investigation, this problem seems to affect multiple pages.

gilesv:
This PR fixes #926.

### CLI page
image

### Networking page
image

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -51,6 +51,11 @@ code {
   border: 1px solid #E1E1E1;
   border-radius: 4px;
   overflow: auto;
+
+  &.block {
+    display: block;
+    white-space: pre;
+  }
 }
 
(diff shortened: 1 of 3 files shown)


greggman:
this is clearly the wrong way to fix this. It's a site wide issue and needs to be fixed at a higher level

JohnTitor:
It seems the issue has been fixed in #932 so we should just close this?

PR #932, opened 2019-10-08
fix code block formatting

greggman:
fixes the code block formatting site wide

See https://github.com/rust-lang/www.rust-lang.org/issues/926

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -53,6 +53,11 @@ code {
   overflow: auto;
 }
 
+pre > code {
+  display: block;
+  white-space: pre;
+}
+
 @mixin code-highlight-white {
   color: white;
   background-color: rgba(black, 0.3);


src/styles/app.scss
```diff
@@ -52,6 +52,19 @@ code {
...
+  line-height: 1.2;
+}
+pre>code {
+  line-height: 1.2;
+  padding: 0;
+  white-space: pre;
+  border: none !important;
+  background: transparent !important;
+  margin: 0;
+}
```
geophree: I think a simpler change would get the same effect.  When I added the following rule to the page in Chrome the issue went away:
```css
pre > code {
  display: block;
  white-space: pre;
}
```

Manishearth:
Should be deployed at http://www-staging.rust-lang.org/ in a bit, please check it out!

LordJuss:
Oh, yes, much better in staging!  Looking forward to seeing it in the actual web-site.

erlend-sh:
Also noticed this on the Networking page and confirmed fixed on staging site.

PR #940, opened 2019-10-29
Deploy translation updates and remove 1aim

skade:
cc @Manishearth


pietroalbini:
r=me once https://github.com/rust-lang/www.rust-lang.org/pull/941 is added

Issue #946, opened 2019-11-07
Move to handlebars-fluent


Manishearth:
This is still somewhat WIP but I published https://github.com/manishearth/handlebars-fluent , which we should be able to use for everything except the `{{team}}` helper. It's directly based off of the i18n code I wrote for this repo.

I'm still documenting it and trying to make it more general, but it's currently in a state which we can use if we desire. Less code for this repo to maintain.

Manishearth:
I'm probably not going to do this yet, but if folks want to help out I can answer any questions you have about the crate (until I add docs)

PR #958, opened 2019-12-06
Move templates closer to what handlebars-fluent uses

### LINKED ISSUE #957: Replace {{text}} with {{fluent}}
Prerequisite for https://github.com/rust-lang/www.rust-lang.org/issues/946, ideally. We can still use it, but we should try and follow the convention that's being created in the fluent ecosystem.

THis also would mean replacing `{{text foo}}` with `{{fluent "foo"}}`

Manishearth:
This bring us closer to #946

r? @Xampprocky or @skade

Fixes #957

### DIFF
--- a/src/i18n.rs
+++ b/src/i18n.rs
@@ -177,14 +177,14 @@ impl HelperDef for I18NHelper {
             id
         } else {
             return Err(RenderError::new(
-                "{{text}} must have at least one parameter",
+                "{{fluent}} must have at least one parameter",
             ));
         };
 
         let id = if let Some(id) = id.path() {
             id
         } else {
-            return Err(RenderError::new("{{text}} takes an identifier parameter"));
+            return Err(RenderError::new("{{fluent}} takes an identifier parameter"));
         };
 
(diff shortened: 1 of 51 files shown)


Manishearth:
After this I think the only work that needs to be done is the actual integration work, the templates are now using the correct format so there won't be much churn.

Manishearth:
I have verified that stuff is still working after this.

XAMPPRocky:
@Manishearth This looks good to me! The one question I have is that you also changed all of the arguments from being like `fluent community-team` to `fluent "community-team"` is this actually important to the code working or is it a style choice?

Manishearth:
Yes, it's important to the code working. See https://github.com/alerque/handlebars-helper-fluent/issues/2 , what we used to do was incorrect. This was mentioned in the second line of https://github.com/rust-lang/www.rust-lang.org/issues/957#

To be able to switch over to handlebars-fluent we need to make this change as well, the old syntax will not work anymore.

XAMPPRocky:
That's fine, I have no preference I just wanted to know for the future.

PR #961, opened 2019-12-10
i18n: only load .ftl files as fluent files

pietroalbini:
While I was working on the localization the application kept panicking at startup, because VIM creates temporary files inside the directory containing the originals, and those files are not valid UTF-8.

This commit changes the localization code to only load `.ftl` files from the `locales/` directory, preventing such issues from happening again.

r? @Manishearth


Manishearth:
Please make this change to the handlebars-fluent crate, we're switching over.

pietroalbini:
Moved it to https://github.com/Manishearth/handlebars-fluent/pull/2.

PR #987, opened 2020-01-23
Fix typo

jieggii:
"билиоитеку" -> "библиотеку"


Manishearth:
This should be done on Pontoon cc @funkill

Issue #586, opened 2018-12-06
Improve legibility of text with low contrast


Aloso:
## Motivation

The W3C recommends for text to use a contrast ratio of at least 4,5:1 for good legibility and at least 7:1 for users with moderately low vision.

This is especially relevant for this website, since because of the flashy colours, the eyes can get tired faster. This website uses text with as little contrast as 4,53 : 1, so it's unsuitable for anyone who does not have perfect eyesight (which includes me):

![contrast ratios](https://user-images.githubusercontent.com/15658558/49616175-33989c80-f9b0-11e8-9ddc-4278fc5fff45.png)

## Solutions

I have some ideas for how this can be improved:

* Make the reddish and greenish colours **slightly darker**
* Add a subtle **text shadow** to improve the contrast

This image is divided into a part without shadow and a part with a shadow (`0 0 5px #00000044`, screenshot from Chromium). The shadow is too subtle to stand out, but it might improve legibility. Of course, making the background darker would be more effective.

![screenshot_20181206_225758](https://user-images.githubusercontent.com/15658558/49616862-81160900-f9b2-11e8-8b30-2617a380420f.png)

* Use black text on a white background for the **majority** of the text. A coloured background could still be used for the header, the footer, for headlines, buttons, icons, etc., but not **continuous text**.

-----

PS: I confused my browser tabs and accidentally filed this issue on blog.rust-lang.org, too. I'm sorry.

ashleygwilliams:
Hi @Aloso ! Thanks for filing- these types of small changes are exactly what we are looking for. I would happily accept a PR to start address this! Feel free to play around with it and we can coordinate in a Work In Progress PR :) Thank you for the thoughtful issue!

Aloso:
@ashleygwilliams Thanks for the response. I [created a PR](https://github.com/rust-lang/www.rust-lang.org/pull/607).

PR #993, opened 2020-01-27
Use a more saturated colour scheme

XAMPPRocky:
This PR changes the colour scheme to be more bright and saturated. This should improve some of the legibility issues.

### Screenshots (Original / New Colours)

Screenshot 2020-01-27 at 21 37 04
Screenshot 2020-01-27 at 21 35 21
Screenshot 2020-01-27 at 21 35 16
Screenshot 2020-01-27 at 21 35 03
Screenshot 2020-01-27 at 21 34 50
Screenshot 2020-01-27 at 21 34 44
Screenshot 2020-01-27 at 21 34 32
Screenshot 2020-01-27 at 21 34 25
Screenshot 2020-01-27 at 21 34 16

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -3,14 +3,14 @@
 $body-font: "Fira Sans", Helvetica, Arial, sans-serif;
 $header-font: "Alfa Slab One", serif;
 
-$gray: #454c52;
-$red: #c14566;
-$green: #398277;
-$purple: #403d58;
-$yellow: #ffd45e;
+$gray: #2a3439;
+$red: #a72145;
+$green: #0b7261;
+$purple: #432866;
+$yellow: #ffc832;
 $border-radius: 4px;
-$link-color: #1eaedb;
-$footer-gray: #303030;
+$link-color: #4299bf;
+$footer-gray: $gray;
 
 html {
   font-size: 62.5%;


pietroalbini:
I don't think this PR goes in the right direction:

* The new green makes accessibility worse, going from a contrast ratio of 4.53 to 3.53.
* The new red seems like a different color than the original.
* The new blue is not really needed, as the old color already has a contrast ratio or 10.36.

Manishearth:
Agreed. Firefox has some a11y devtools that help with contrast.

Also a common complaint about the website styling was that it was very bright, making it more saturated doesn't help.

pietroalbini:
Tried to improve the colors without changing the overral look of the website:

* For the red, `#ad405d` goes from `4.87` to `5.72`.

  

* For the green, `#306e65` goes from `4.53` to `5.93`.

XAMPPRocky:
Okay, I've opted for more deeper colours, and ensured all of this colour scheme passes contrast ratios tests on all pages.

@pietroalbini Once a a contrast is above 4.5:1, having higher contrast above that isn't an empiric good, there's a lot of other factors consider beyond contrast ratio.

@Manishearth That kind of feedback isn't very useful for me, as it is not actionable. There's no time to that complaint, so I have nothing to compare it to, the colours have not remained the same since the redesign. Saturation is not the same as brightness, and a complaint about the website being too bright can be about a lot more than the colours being too bright. At the end of the day a lot of design is going to be subjective, and I'm going to make the changes I believe are for the better, they will never satisfy everyone.

Screenshot 2020-01-28 at 13 43 07
Screenshot 2020-01-28 at 13 43 02
Screenshot 2020-01-28 at 13 42 46
Screenshot 2020-01-28 at 13 42 33
Screenshot 2020-01-28 at 13 42 24

Manishearth:
@XAMPPRocky it's not any less actionable than this pull request: I don't want to merge a design change with very little justification when there were also complaints in the opposite direction.

Yes, brightness and saturation aren't the same technical term, but when used colloquially they often are. And yes, the colors have changed since then, but a general desire to avoid very bright colors still makes sense.

----

The deeper colors do look better and they don't look "bright".

pietroalbini:
> Once a a contrast is above 4.5:1, having higher contrast above that isn't an empiric good, there's a lot of other factors consider beyond contrast ratio.

@XAMPPRocky The issue the PR you closed was addressing, https://github.com/rust-lang/www.rust-lang.org/issues/586, mentioned that the reporter had trouble reading the website even with the 4.5:1 contrast ratio due to their low vision, and they suggested a higher ratio.

---

For the rest of the stuff I agree with Manish.

XAMPPRocky:
Okay, I've opted for a slightly deeper green. Here's the current contrast ratios for white foreground on colour background. @pietroalbini For what it's worth this PR's contrast ratios are higher than the colours purposed in the reporter's fork.

* red — [7.12](https://color.a11y.com/ContrastPair/?bgcolor=ffffff&fgcolor=c14566)
* green — [5.84](https://color.a11y.com/ContrastPair/?bgcolor=ffffff&fgcolor=c14566)
* purple — [12.11](https://color.a11y.com/ContrastPair/?bgcolor=ffffff&fgcolor=c14566)

Screenshot 2020-01-29 at 11 57 50
Screenshot 2020-01-29 at 11 58 00
Screenshot 2020-01-29 at 11 58 26
Screenshot 2020-01-29 at 11 58 32

pietroalbini:
Can we still try to keep the new colors as close as possible to the original ones, while increasing the accessibility?

Manishearth:
I like the deeper colors. I don't have strong preferences either way on old vs new, but if this helps a11y that's good. I'm r+ for that change but I'd like to wait for Pietro to be r+ as well.

pietroalbini:
The only color I'm not too convinced about is the blue/purple... Could you keep it a little closer to the original one?

XAMPPRocky:
@pietroalbini If this is purely a preference, I would prefer to retain my agency to make these kinds of subjective decisions while I'm the active maintainer, and keep the purposed colours as is. I don't see a reason to be precious about the current colour scheme, or this one, or any other one. I don't think we've ever made a promise to commit to a particular colour scheme and I don't see why we would now. It will never be perfect, and we'll continue to improve it over time.

Manishearth:
Uh, pietro and I have also been actively maintaining the website.

I don't really have strong opinions about colors, but "let's not be precious about a color scheme" cuts both ways.

XAMPPRocky:
@Manishearth I didn't mean to imply that you weren't. I think if I were being being precious I wouldn't have *already* made a number of changes, and I think if we're at the point of discussing preference I would want to get feedback from a broader set of people and the only reasonable way to get that feedback is to deploy and test it. And in the future I do want to test and try out radically different colour schemes, however that takes a lot longer to figure out This is meant as a small incremental improvement to the current scheme in the mean time.

Zexbe:
The current color scheme reminds me of fog. It looks okay, but it is harder to read. The new theme looks better, but I think the "build it in rust" would look better if it went in the other direction, and had the background much darker to get the contrast. I think not involving some preference is going to be impossible, if the great light vs dark theme battle is any indicator. 


Darker "Build it" Section

build-it-in-rust

Alternate colors I tried. (I only changed enough to get the idea across)

cool colors

XAMPPRocky:
@Zexbe The reason it looks like fog is because all of the colours are very light and when paired with a white backgrounds in other sections it looks like colours are washed out. The colours in the PR are essentially the original colours with the lightness turned down and their saturation increased. I do like the black/dark navy for "Build It In Rust" though I'm not sure it fits with the other colours.

Since the discussion has slowed down, I don't want this PR to become derailed into discussing the colour scheme in general, and I want to close the contrast issues I'm going to merge this. We just deployed so this will have a decent amount of amount of time on staging so people can try it out and if anyone finds an erroneous case we can fix/revert it.

This isn't a commitment to these particular colours, we can and probably will change them again in the future. Personally, I would be okay with having "seasonal" (read some period of months) colour changes.

pietroalbini:
> Since the discussion has slowed down, I don't want this PR to become derailed into discussing the colour scheme in general, and I want to close the contrast issues I'm going to merge this.

If the intention is to fix the accessibility problem then we should just improve the less accessible colors, and properly discuss the other color changes in another PR, without fast-tracking it.

PR #994, opened 2020-01-27
Add syntax highlighting

XAMPPRocky:
This PR adds syntax highlighting to the website using `highlight.js`.

### Screenshot (Current / Highlighted) — Current Colours
Screenshot 2020-01-27 at 22 32 36

### Screenshot (Current / Highlighted) — #993 Colours
Screenshot 2020-01-27 at 22 32 16


XAMPPRocky:
@Manishearth Would you mind giving this PR a review?

Manishearth:
Eventually. I don't always have 24hr turnaround time on review, sorry, and right now I'm at the Mozilla all hands.

XAMPPRocky:
@Manishearth That's totally understandable. I think for the future it wouldn't be fair with me taking a more active lead and maintenance of the website, for you to be required to review every commit. I think it would be more flexible to use master as push, and then once a week I'll make a deployment PR from `master -> deploy`, and leave that for your review? It'll be a lot easier for you to review as well as you can easily compare staging to production.

Manishearth:
Uh, guaranteed sub-24 hour review cycles aren't the norm _anywhere_ in the rust project. I'm also not the only reviewer (Pietro and Florian are both around).

Let's continue to have a pristine master, so that deploys can be very low-effort. I will try to help review as much as possible, I just can't guarantee very fast turnaround time, which should be fine.

Manishearth:
Seems like it's using automatic language detection: Should we explicitly annotate with the language for as many codeblocks as possible?

XAMPPRocky:
@Manishearth Well I don't think it will be a problem as the only languages included in the pack are Rust, Markdown, TOML, Bash, JSON, YAML, and plaintext. So Rust code is pretty unlikely to be misdetected. For future proofing though we should start to annotate them.

Manishearth:
Yeah precisely.

PR #1000, opened 2020-02-01
Remove white from code background

XAMPPRocky:
While looking into another issue I noticed the code snippets have a white background that makes them impossible to see, this PR fixes that.
Screenshot 2020-02-01 at 10 36 28


pietroalbini:
This looks good, but can you provide screenshots for the other backgrounds?

XAMPPRocky:
@pietroalbini This removes the background colour. So all of them will look the same as regular text.
Screenshot 2020-02-01 at 11 30 59

Issue #969, opened 2019-12-23
Difficulty identifying install button as a button (Windows)


linhub15:
# Design Feature Request
## Summary
The install button is hard to identify as a clickable button (https://www.rust-lang.org/tools/install).
Should either be smaller or use a verb.

![image](https://user-images.githubusercontent.com/10420994/71372184-e95d6b80-2570-11ea-845d-5515748b3520.png)

### Using a verb
![image](https://user-images.githubusercontent.com/10420994/71373258-ca140d80-2573-11ea-92af-c7fc76216ae6.png)

## Motivation
It is difficult to immediately identify it as a button. It passes off as another piece of content, like an aside, card or section.

Also, we should be consistent with button text. Other site buttons use verbs and/or have smaller widths to easily identify as buttons.

![image](https://user-images.githubusercontent.com/10420994/71373146-6b4e9400-2573-11ea-89ed-803f497cbc88.png)

## Drawbacks
* No drawbacks by adding a verb
* If adding `max-width` then decisions will need to be made for alignment, and `max-width` amount.

## Rationale and alternatives
- Ideally it would be nice to have a seamless install for those that are seeking to get started with Rust. The impact of not doing this is that **some** Windows users **may** struggle for a few seconds when trying to install Rust. 
- This is the best design because it is consistent with the rest of the site
- Alternative: give `.buttons` class a `max-width` so it will not span the entire width of content
![image](https://user-images.githubusercontent.com/10420994/71373288-e44deb80-2573-11ea-9bc4-bd924a33271a.png)
- Alternative 2: Use a verb and add `max-width`

## Maintenance
N/A

## Unresolved Questions
N/A

Issue #954, opened 2019-12-02
Was given the Unix instruction not the Windows ones.


Y-Less:
# Content Bug Report

## Page(s) Affected
https://www.rust-lang.org/tools/install

## What needs to be fixed?

Windows download link in not on the getting started page.  The website detected me as on Unix, and didn't provide any fallback option to say "no I'm not":

![image](https://user-images.githubusercontent.com/991078/69952519-a2430400-14f7-11ea-8d4f-505782bacd9a.png)

There is a "Other installation methods" link, but it is a 404.  I eventually found a link to the Windows installer in the rustp repo, but it wasn't obvious.

## Suggested Improvement
(optional) Do you have a suggestion for how to fix the problem? Include your ideas
here.

steveklabnik:
What OS (with version) are you on? What browser are you using?

Y-Less:
Windows 8.1.  Waterfox 56.2.4 (64-bit)

The issue was less that the website got it wrong, and more that there's no way to correct it (which also means there's no way to download for a different computer).

I also just tested on Waterfox 2019.10 and got the same thing.

skade:
Seems like Waterfox only reports itself as Win64 and we only check for Win32.

https://github.com/rust-lang/www.rust-lang.org/blob/master/static/scripts/tools-install.js#L13-L57

I'm wondering if we're even bugged on other platforms and only saved by the fallback triggering on all major browsers.

steveklabnik:
> and more that there's no way to correct it (which also means there's no way to download for a different computer).

At the bottom of the page, there's a link to https://forge.rust-lang.org/infra/other-installation-methods.html, which shows you installation instructions for all platforms. Well, it *should* link to that, but doesn't, that's https://github.com/rust-lang/www.rust-lang.org/issues/955

Issue #733, opened 2019-03-28
Windows installation instructions


dcarosone:
# Content Bug Report

## Page(s) Affected
Getting started / installation 

~https://www.rust-lang.org/learn/get-started~
https://www.rust-lang.org/tools/install

## What needs to be fixed?

I wanted to look at the instructions for installing on Windows the other day, because i remembered that there were a bunch of extra steps and dependencies. I don't normally use windows, but I was curious about the viability of installing it on a customer workstation for a quick task.

So I went to the website, but I couldn't even *see* the Windows install instructions unless I changed my user-agent string to pretend to be on Windows.  And even then, all I was told was "get rustup-init.exe and follow the on-screen instructions".

I just want to review those instructions!

Following through a series of "learn more" type links led me on a path through these pages, neither of which contained the information needed.
* https://www.rust-lang.org/tools/install
* https://forge.rust-lang.org/other-installation-methods.html

I'm sure this used to be visible previously.
**Edit**: it's in the install page, but hidden for non-windows users, with no noticable way of exposing it.

## Suggested Improvement

Provide users a way to view instructions for platforms other than what's auto-detected, maybe via an expanding section or tabbed panel (have the detection pick the relevant tab, rather than completely hide other content). Or just don't hide this paragraph for anyone.

~Provide detail about platform steps and additional dependencies in the online installation instructions, rather than (only) in the installation tool.~

steveklabnik:
Here's the old page: https://prev.rust-lang.org/en-US/install.html

I think that everything that used to be visible is the exact same on the new page.

dcarosone:
Thanks, this (and the link to the additional detail) was precisely what I remembered and was looking for:

![image](https://user-images.githubusercontent.com/11495017/55199353-b9d02d00-51b1-11e9-9d47-a223f1960267.png)

It's not in the main page. It turns out, it *is* in the second page (tools/install) behind the "learn more" banner link, *when I come from a browser on Windows*.  It's there in the source, with some class and style stuff to hide it - apparently my user-agent tricks didn't work well enough.

So, the request reduces to the first part: being able to find this information from other platforms: something that can be collapsed by default but visible and expanded if you want. 

We (effectively) show the linux instructions to Windows users for WSL anyway, so there might be some nice way to generalise this selection/navigation.

dcarosone:
(or, frankly, since it's only a heading and single paragraph, already behind a would-you-like-to-know-more? link, just don't hide this at all)

[...]

PR #1004, opened 2020-02-03
Make install button easier

Zexbe:
Fixes: https://github.com/rust-lang/www.rust-lang.org/issues/969
Possibly Fixes: https://github.com/rust-lang/www.rust-lang.org/issues/954
Possibly Fixes: https://github.com/rust-lang/www.rust-lang.org/issues/733

### **Before**
Before

### **After**
mini
small
larger

### DIFF
--- a/locales/en-US/tools.ftl
+++ b/locales/en-US/tools.ftl
@@ -60,6 +60,10 @@ tools-install-page-title = Install Rust
 
 install-using-rustup-heading = Using rustup (Recommended)
 
+install-rustup32-button = Download <nobr>rustup-init.exe</nobr> <nobr>(32-bit)</nobr>
+
+install-rustup64-button = Download <nobr>rustup-init.exe</nobr> <nobr>(64-bit)</nobr>
+
 install-notes-heading = Notes about Rust installation
 
 install-notes-getting-started-description = If you're just getting started with
@@ -149,8 +153,8 @@ install-other-methods-link = Learn more
 
 ## components/tools/rustup.hbs
 
-tools-rustup-unixy = It looks like you’re running macOS, Linux, or another Unix-like OS. To download Rustup and install Rust, run the following in your terminal, then follow the on-screen instructions.
-tools-rustup-windows = It looks like you’re running Windows. To install Rust, download and run the following, and then follow the onscreen instructions.
+tools-rustup-unixy = It looks like you’re running macOS, Linux, or another Unix-like OS. To download Rustup and install Rust, run the following in your terminal, then follow the on-screen instructions. If you are using Windows, <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">click here</a>
+tools-rustup-windows = It looks like you’re running Windows. To start using Rust, download the installer, then run the program and follow the onscreen instructions. If you are not using Windows, <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">click here</a>
 tools-rustup-wsl-heading = Windows Subsystem for Linux
 tools-rustup-wsl = If you’re a Windows Subsystem for Linux user run the following in your terminal, then follow the on-screen instructions to install Rust.
 tools-rustup-unknown = Rust runs on Windows, Linux, macOS, FreeBSD and NetBSD. If you are on one of these platforms and are seeing this then please report an issue with the following values:
--- a/locales/en-US/tools.ftl
+++ b/locales/en-US/tools.ftl
@@ -60,9 +60,9 @@ tools-install-page-title = Install Rust
 
 install-using-rustup-heading = Using rustup (Recommended)
 
-install-rustup32-button = Download <nobr>rustup-init.exe</nobr> <nobr>(32-bit)</nobr>
+install-rustup32-button = Download <span class="nowrap">rustup-init.exe</span> <span class="nowrap">(32-bit)</span>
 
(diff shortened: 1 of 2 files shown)


pietroalbini:
Thanks for the PR!

As far as I'm aware this would be the only page with an 80% width paragraph, which is not consistent with the rest of the site. An alternative approach would be to keep the 100% width for the paragraph while having two buttons instead of one in the row below:

* [Download rustup-init.exe (32-bit)](https://static.rust-lang.org/rustup/dist/i686-pc-windows-msvc/rustup-init.exe)
* [Download rustup-init.exe (64-bit)](https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe)

This would also provide yet another band-aid for https://github.com/rust-lang/rustup/issues/2179 (cc @kinnison).

Also, do not update the translations by changing the ftl files directly in the repo: the English language is the only one supposed to change, the other ones need to be updated through Pontoon.

Zexbe:
Okay, ready to be reviewed

pietroalbini:
This looks way better! Thanks!

Just a last small change, could you add some margin between the buttons and the following title?

Zexbe:
Okay

locales/en-US/tools.ftl
```diff
@@ -149,8 +153,8 @@ install-other-methods-link = Learn more
 
 ## components/tools/rustup.hbs
 
-tools-rustup-unixy = It looks like you’re running macOS, Linux, or another Unix-like OS. To download Rustup and install Rust, run the following in your terminal, then follow the on-screen instructions.
-tools-rustup-windows = It looks like you’re running Windows. To install Rust, download and run the following, and then follow the onscreen instructions.
+tools-rustup-unixy = It looks like you’re running macOS, Linux, or another Unix-like OS. To download Rustup and install Rust, run the following in your terminal, then follow the on-screen instructions. If you are using Windows, <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">click here</a>
+tools-rustup-windows = It looks like you’re running Windows. To start using Rust, download the installer, then run the program and follow the onscreen instructions. If you are not using Windows, <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">click here</a>
```
pietroalbini:
```suggestion
tools-rustup-windows = It looks like you’re running Windows. To start using Rust, download the installer, then run the program and follow the onscreen instructions. If you are not using Windows, <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">click here</a>.
```

locales/en-US/tools.ftl
```diff
@@ -149,8 +153,8 @@ install-other-methods-link = Learn more
 
 ## components/tools/rustup.hbs
 
-tools-rustup-unixy = It looks like you’re running macOS, Linux, or another Unix-like OS. To download Rustup and install Rust, run the following in your terminal, then follow the on-screen instructions.
-tools-rustup-windows = It looks like you’re running Windows. To install Rust, download and run the following, and then follow the onscreen instructions.
+tools-rustup-unixy = It looks like you’re running macOS, Linux, or another Unix-like OS. To download Rustup and install Rust, run the following in your terminal, then follow the on-screen instructions. If you are using Windows, <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">click here</a>
```
pietroalbini:
```suggestion
tools-rustup-unixy = It looks like you’re running macOS, Linux, or another Unix-like OS. To download Rustup and install Rust, run the following in your terminal, then follow the on-screen instructions. If you are using Windows, <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">click here</a>.
```

XAMPPRocky: Don't use "click here" for links, it's bad content. The link text should describe what you're clicking on.
```suggestion
tools-rustup-unixy = It looks like you’re running macOS, Linux, or another Unix-like OS. To download Rustup and install Rust, run the following in your terminal, then follow the on-screen instructions. See <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">"Other Installation Methods"</a> if you are on Windows.
```

locales/en-US/tools.ftl
XAMPPRocky: Same comment as above.

Zexbe:
Okay, updated the text

locales/en-US/tools.ftl
```diff
@@ -60,6 +60,10 @@ tools-install-page-title = Install Rust
 
 install-using-rustup-heading = Using rustup (Recommended)
 
+install-rustup32-button = Download <nobr>rustup-init.exe</nobr> <nobr>(32-bit)</nobr>
```
XAMPPRocky: `nobr` is [non-standard](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nobr). 
```suggestion
install-rustup32-button = Download <span class="nowrap">rustup-init.exe</span> <span class="nowrap">(32-bit)</span>
```

locales/en-US/tools.ftl
```diff
@@ -60,6 +60,10 @@ tools-install-page-title = Install Rust
...
+
+install-rustup64-button = Download <nobr>rustup-init.exe</nobr> <nobr>(64-bit)</nobr>
```
XAMPPRocky:
```suggestion
install-rustup64-button = Download <span class="nowrap">rustup-init.exe</span> <span class="nowrap">(64-bit)</span>
```

XAMPPRocky:
Okay, this looks looks good to me! Thank you again for your PR!

PR #1009, opened 2020-02-11
Add support for running the page behind a proxy.

ericho:
This is a non-important issue, but it might be helpful for some
people.

The rust web page fetch data from the rust-lang.org site
when it loads. If the site is run locally and behind a proxy,
this fetch process will fail with timeout error. `reqwest` needs
to know explicitly the information about the proxy.

This patch add a function to fetch this data by retrieving the
proxy information from the environment variables `https_proxy` and
`HTTPS_PROXY`. Then the `reqwest` client is created accordingly.

A minor refactor was done to call this new function from the
`fetch_rust_*` functions.

Signed-off-by: Erich Cordoba <erich.cm@yandex.com>

### DIFF
--- a/src/rust_version.rs
+++ b/src/rust_version.rs
@@ -1,21 +1,50 @@
 use reqwest;
 use serde_json;
 use std::any::Any;
+use std::env;
 use std::error::Error;
 use toml;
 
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Yep, looks good to me. Thank you for your PR, and congrats on your first contribution! 🎉

PR #1010, opened 2020-02-14
Update footer attribution

XAMPPRocky:
This changes the language to be more general, updates the link the new issue page, and removes the "click here", with an active description


Manishearth:
Please update the URLs for all languages.

XAMPPRocky:
@Manishearth Done, also tangentally what is `xx-AU`, and why do we have it? It's the text upside down?

Manishearth:
It's a testing locale, mostly used when messing around with localization. It's easy to see if stuff is not localized when everything that is localized is upside down. It's not actively updated, it's mostly there for when we need to test things.

PR #1011, opened 2020-02-14
update production user mail address to core (for now)

llogiq:
Hopefully this will avoid further confusion until a new mail address is set up.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,7 +37,7 @@ page of the site.
 
 If you would like your company to be displayed on the 'Production' page
 (i.e. would like to edit `src/data/users.yml`), you should
-e-mail production@rust-lang.org (please do *not* raise a PR) with:
+e-mail core@rust-lang.org (please do *not* raise a PR) with:
 
  - Your company name, URL and logo.
  - A one short sentence describing your usage of Rust, ideally with a link


Manishearth:
cc @nikomatsakis i presume this is what we want for now?

llogiq:
At least that's what I was advised to do on Twitter when I asked about it.

XAMPPRocky:
The current instructions don't make sense to me. I would prefer if people just opened a PR, I don't know why we have to put it in. Core is not going to make the PRs, and I would prefer to avoid just getting forwarded tasks in email.

llogiq:
Fine with me. So I'll remove the  section instead & send another PR to add my company?

Manishearth:
No, there needs to be an email sent so that we have a contact (which shouldn't be required to be on a public repo). The production WG was intended to help manage that, but in lieu of that we should still require the email to be sent. We can change the instructions to ask them to open a PR _and_ send an email to core.

llogiq:
I already sent the email, so...I'll take that as a yes. :-)

XAMPPRocky:
@Manishearth Are we tracking these contacts somewhere? I'm not aware it if so.

Manishearth:
There was a spreadsheet. But even having it within core team/production team's inboxes for later retrieval is a step.

XAMPPRocky:
@Manishearth We talked about setting up a website team, if we that made on `rust-lang/team` we could have a `website@` mailing list, and ask people to email that instead so it gets sent to us.

Manishearth:
The goal of maintaining this list is to contact them occasionally for production user outreach, which doesn't quite fall within the purview of the website team. We should ideally have a team for doing this, which previously didn't work out because key people were busy, but might once we have the newer WG model.

It might make sense as part of the community team. The current Excel sheet contains some private company names (who haven't publicly talked about their Rust usage), so it can't easily be opened to a wider team, but for things being put on the website, this could potentially be done.

XAMPPRocky:
Alright, I'm going to to merge this for now, to remove the out dated info from the website, and we can figure out how best to handle production users in seperate issue.

PR #1017, opened 2020-02-21
Add Terramagna in production users

landreussi:
We are using Rust in many services at Terramagna, we really like it and we want to put our name in the rust users page.

Captura de tela de 2020-02-21 15-23-04


XAMPPRocky:
Thank you for your PR, and congrats on your first contribution! 🎉

PR #1018, opened 2020-02-29
Fix where to talk section buttons to be properly aligned

### LINKED ISSUE #918: Buttons in `Where to talk` section on the Community page are not aligned
# Design Bug Report

## Page(s) Affected
Buttons in the `Where to talk` section on the community page are not aligned. They should be either top aligned or bottom aligned. Seems previously in #551 it was requested to be top-aligned. Looks like a combo of #793 and #872 caused the alignment to be off.

## What needs to be fixed?
To top align the buttons, we would just remove the `flex-grow-1` on the items in this section. This would currently work as all the `divs` have same number of lines of content. We could add a `height` (`250px` for 1 line, `290px` for 2 lines) to push them down further for a little buffer.

To bottom align the buttons, we would just remove the `mb3` class on the `li` element surrounding the `Learn more about teams` button.

## Suggested Improvement
I have no opinion on which way they should align. I would love to do the PR to fix it once a decision has been made.

Attaching some screenshots for clarity.

### Current

image

### Bottom aligned

<details>

</details>

### Top aligned (with room for 1 more line)

<details>

</details>

### Top aligned (with room for 2 more lines)

<details>

</details>

wyze:
Fixes #918

I went a little further and ensured that when the content has different lengths, the buttons will still align properly. This did result in duplicating the button content though.

image

I'm more than happy to revert this and go back to my original approach, but when the content length doesn't match, the buttons will no longer be aligned.

### DIFF
--- a/templates/community/index.hbs
+++ b/templates/community/index.hbs
@@ -37,37 +37,62 @@
       <div class="highlight"></div>
     </header>
     <div class="flex flex-column flex-row-l">
-      <div class="mw-33-l mh3-l pt0 flex flex-column justify-start" id="users-forum">
-        <div class="flex-grow-1">
+      <div class="w-33-l mh3-l pt0 flex flex-column justify-start" id="users-forum">
+        <div>
           <h3 class="mt0 mb4">{{fluent "community-urlo-header"}}</h3>
           <p>
             {{fluent "community-urlo"}}
           </p>
         </div>
 
-        <a href="https://users.rust-lang.org" class="button button-secondary">{{fluent "community-discourse-button"}}</a>
+        <span class="dn-l">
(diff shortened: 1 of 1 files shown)


Manishearth:
Can you check that it looks good on all the translations as well?

wyze:
That is a good point. I much prefer that I did go the route I did as previously I didn't even think about other translations.

For instance, here is Portuguese in about the same screen width:

image

Whereas my original approach wouldn't have worked, see here:

image

Manishearth:
Right, the `flex-grow-1` was introduced specifically because the styling was broken for some translations, so I suspected that a change _just_ reverting that would not work well

Manishearth:
Thanks!

PR #1020, opened 2020-03-01
Use different purple

XAMPPRocky:
This updates the colour scheme based on talking with @pietroalbini to use a purple that is more deep blue.

- **Try It Out** — https://www-try.rust-lang.org
- **Staging** — https://www-staging.rust-lang.org
- **Production** — https://www.rust-lang.org

### New / Production

Screenshot 2020-03-01 at 14 09 58

### New / Master

Screenshot 2020-03-01 at 14 10 18

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -6,7 +6,7 @@ $header-font: "Alfa Slab One", serif;
 $gray: #2a3439;
 $red: #a72145;
 $green: #0b7261;
-$purple: #432866;
+$purple: #2e2459;
 $yellow: #ffc832;
 $border-radius: 4px;
 $link-color: #4299bf;


pietroalbini:
I'd slightly prefer a saturation of 30% / 35%, but this looks way better to me! @Manishearth

XAMPPRocky:
Here's a screenshot of 30%, personally I don't like it as it looks too gray compared to the new purple.

### New / 30% Saturation

Screenshot 2020-03-01 at 20 49 50

pietroalbini:
The color introduced in the PR is starting to grow on me! Let's wait for @Manishearth's approval and then we can merge this.

By the way, I chatted with Erin and it turns out there was a miscommunication, the color with 30% saturation I was proposing is `#312a4f`.

Manishearth:
Yeah, this looks way better to me as well (and my understanding is that in the poll about colors people liked the new green but not the new-now-old purple)

PR #1025, opened 2020-03-05
Display team alumni in the governance page

pietroalbini:
This PR implements support in the website for "team alumni", in the effort to move away from the alumni page. This PR displays alumni under each team instead, like this screenshot with dummy data:

At the moment no actual change will be shown, as nobody is marked as alumni in the team repo yet.

r? @XAMPPRocky

### DIFF
--- a/templates/governance/group-team.hbs
+++ b/templates/governance/group-team.hbs
@@ -52,5 +52,26 @@
                 </div>
             {{/each}}
         </div>
+        {{#if team.alumni}}
+            <header class="pb3 pt3">
+                <h2 class="f2">{{fluent "governance-alumni-header"}}</h2>
+            </header>
+            <p>{{fluent "governance-alumni-thanks"}}</p>
+            <div class="flex flex-column flex-row-l flex-wrap-l justify-start">
+                {{#each team.alumni as |member|}}
+                <div class="w-100 w-33-l mb3 flex flex-row items-center">
+                    <a class="mr4 w3 h3 flex-no-shrink" href="https://github.com/{{member.github}}">
+                        <img class="w-100 h-100 bg-white br2" src="https://avatars.githubusercontent.com/{{member.github}}">
+                    </a>
+                    <div>
+                        {{member.name}}
+                        <div class="f4">
+                            GitHub: <a href="https://github.com/{{member.github}}">{{member.github}}</a>
+                        </div>
+                    </div>
+                </div>
+                {{/each}}
+            </div>
+        {{/if}}
     {{/if}}
 </div>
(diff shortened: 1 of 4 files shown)


XAMPPRocky:
This looks good. Could you also include a screenshot of how this looks on mobile?

pietroalbini:
On mobile:

Screen Shot 2020-03-05 at 16 16 54

Also fixed the tests.

XAMPPRocky:
Hmm, on mobile I think it might look better if it was two people per row.

pietroalbini:
That could break with long names (such as Aidan Hobson Sayers).

templates/governance/group-team.hbs
```diff
@@ -52,5 +52,23 @@
...
         </div>
+        {{#if team.alumni}}
+            <header class="pb3 pt3">
+                <h2 class="f2">{{fluent "governance-alumni-header"}}</h2>
+            </header>
+            <p>{{fluent "governance-alumni-thanks"}}</p>
+            <div class="flex flex-column flex-row-l flex-wrap-l justify-start">
+                {{#each team.alumni as |member|}}
+                <div class="w-100 w-33-l mb3 flex flex-row items-center">
+                    <a class="mr3 w2 h2 flex-no-shrink" href="https://github.com/{{member.github}}">
```
XAMPPRocky: Okay I tested this out locally, and I think if the images were slightly bigger overall it would better for desktop but particularly mobile.

Screenshot 2020-03-07 at 12 21 25


```suggestion
                    <a class="mr3 w3 h3 flex-no-shrink" href="https://github.com/{{member.github}}">
```

pietroalbini: I'm not sure if that's too much emphasis.

XAMPPRocky: I disagree that it's putting too much emphasis, it's takes up a quarter of the area that current member's image does, that's a pretty clear visual hierarchy, I don't think you could ever confuse the two sections.

For me a lot of the images are illegible at the current size. This change makes the image a similar size to the avatar in a top level GitHub comment, which is where most people see their avatar, so it's more likely to be legible.

I also don't think I agree with the motivation that alumni should not have too much emphasis. It needs to be clearly distinguished from the current members sections, but not necessarily de-emphasised. These people's work was important, we shouldn't de-emphasise that.

pietroalbini: Hmm, the big avatar with the small name doesn't look too great... We could add the GitHub username as well, but it emphasizes the "GitHub label" problem even more:

XAMPPRocky: I like that approach. It's more consistent. I can make a PR later switching the GitHub label to an image.

pietroalbini: Ok updated the PR.

PR #1029, opened 2020-03-10
Rename rustc-guide to rustc-dev-guide

chrissimpkins:
The rustc-guide repository was moved to rustc-dev-guide and the new URL for the rustc-dev-guide is https://rustc-dev-guide.rust-lang.org/

The discussion about these changes is in https://github.com/rust-lang/rustc-dev-guide/issues/470.

This PR revises `rustc-guide` references in this repository to the new name.

Transition tracker: https://github.com/rust-lang/rustc-dev-guide/issues/602


chrissimpkins:
The transition is complete and this PR is ready for review/merge. Thanks!

JohnTitor:
I think translation will be done on pontoon side, so you don't have to update `locales/ja/teams.ftl` here.

pietroalbini:
Yep the translations have to be done through Pontoon.

chrissimpkins:
Reverted ja translation in https://github.com/rust-lang/www.rust-lang.org/pull/1029/commits/00f3413cac0071dca4c67ce50f22e6b73ff5226a

XAMPPRocky:
I think there's some confusion here. The repository is the canonical source of the translations, and pontoon syncs to the repository, but changes aren't required to go through pontoon. In general, changes to translations should go through pontoon so that they go through proper review. But cases like this where it is the same word across all translations it's preferable to make the change in the PR to reduce the overall work.

@chrissimpkins Would you mind re-adding the japanese localisation and any others that are present?

chrissimpkins:
In only came across `en-US` and `ja`.

I spot checked other localizations and it looks like the translations are using some form of "rustc guide" / "Rustc Guide" translation for the name.  For instance, in `it` it is: "la guida rustc".  Do we want to abandon this approach and fix the name as `rustc-dev-guide` or push translations through Pontoon with a new name "Rustc Dev Guide"?

cc @mark-i-m @spastorino

mark-i-m:
Hmm... my preference is to uniformly call it `rustc-dev-guide` everywhere, even changing it in English from "rustc guide" or "rustc dev guide" or variants.

pietroalbini:
> I think there's some confusion here. The repository is the canonical source of the translations, and pontoon syncs to the repository, but changes aren't required to go through pontoon.

To be honest I didn't work much with the translations, but I thought changes were required to go through it. Oh well, everyone learns something every day :)

chrissimpkins:
>Hmm... my preference is to uniformly call it `rustc-dev-guide` everywhere, even changing it in English from "rustc guide" or "rustc dev guide" or variants.

@mark-i-m Any idea how we do this? Is it sufficient to add this to the en-US and push to Pontoon or do we need to translate from localized writing systems?

XAMPPRocky:
Looking into this further, this PR is not actually needed at all. Since this info is autogenerated from the team repo. The current team repo doesn't mention rustc-guide at all, it's just that team data is quite out of date. I've filed #1030 to address that.

PR #1030, opened 2020-03-10
Update Team Data

XAMPPRocky:
This updates the team data based on `cargo run dump-website` in https://github.com/rust-lang/team


upsuper:
Hey @XAMPPRocky @Manishearth! This commit was problematic for translation, because it updates the content of some strings without changing the string key, for example the value of `governance-team-wg-rls-2-name` was changed from `RLS 2.0` to `RLS 2.0 working group`.

Pontoon doesn't seem to be able to detect this kind of changes, and would assume that existing translations for the old strings can still be used, until someone happens to notice the mismatch.

We should probably have some mechanism to check against updating strings without changing the key accidentally. WDYT?

upsuper:
Raised an issue in #1091.

PR #1032, opened 2020-03-11
Roll Japanese Translations

vertexclique:
Adds Japanese translation deployment, checked it locally. Looks like it is translated mostly.
Let's give it a try.

@rust-lang/community-localization


XAMPPRocky:
Looking at pontoon, the only thing that isn't translated is the legal policy which seems fine if it comes later if at all.

PR #1033, opened 2020-03-11
Deploy Master Changes

XAMPPRocky:


Manishearth:
Given that this has two new languages, we should ask the localizers to verify that the staging site is working correctly before merging.

Manishearth:
@vertexclique can you ping the relevant translation teams and ask them to have a look? Not necessarily a super thorough one

XAMPPRocky:
@Manishearth Both of the leads on those translation teams reviewed it earlier today said it looked good.

Manishearth:
Gotcha. Go for it then.

Just a note for future deploys: make sure you cycle through a bunch of pages on the site; the reason the staging site exists is to make sure we don't accidentally break everything.

(Currently the site seems good)

XAMPPRocky:
@Manishearth What makes you think I'm not doing that?

Manishearth:
Nothing, I just wasn't sure if I'd mentioned this earlier and wanted to ensure we're still doing deploys properly.

PR #1035, opened 2020-03-12
Localize Privacy Notice

### LINKED ISSUE #1034: Localize privacy policy
### What needs to be fixed?
The Privacy Policy page is not localizable.

### Page(s) Affected
https://www.rust-lang.org/es/policies/privacy

### Suggested Improvement
We should move the content to a fluent file. I have a PR ready!

jrvidal:
Fixes #1034

### DIFF
--- a/locales/en-US/privacy.ftl
+++ b/locales/en-US/privacy.ftl
@@ -3,3 +3,174 @@
 ## templates/policies/privacy.hbs
 
 policies-privacy-page-title = Rust Privacy Notice
+
+policies-privacy-page-version =
+    <p>Version { $version }, dated <time datetime="{ $date }">{ $date }</time></p>
+
+policies-privacy-page-intro =
+    <p>
+      The Rust Project oversees a number of projects, including the
+      <a href="{ $baseurl }/">Rust website</a>,
+      <a href="https://crates.io">crates.io</a>, and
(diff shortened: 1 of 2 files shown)


Manishearth:
Thanks! Once this gets deployed to www-staging.rust-lang.org/ I'll have a closer look to see if there are any differences.

XAMPPRocky:
@Manishearth Should we have something at the top saying the english version is the canonical version like people have suggested for other policies?

Manishearth:
Oh, yes, definitely. Good catch

Manishearth:
The content hasn't changed :+1: We're good to go for translating it, though we should still add that header.

upsuper:
Hey! Just noticed this new string now. Can we avoid put large chunk of HTML like `policies-privacy-page-crates-io-desc` into text piece? Pontoon's UI is not particularly optimized for large chunk of text like this. It's hard to translate (you have to translate all in one go or nothing), hard to review (you have to manually scroll original and translation), hard to distribute the work (again, you have to do them altogether), hard to update (you can't reuse anything from previous translation even if only minor changes are done to the text).

It would be appreciated if this kind of text can be broken down to just one item a string instead, e.g. rather than having `<dl><dt>` all in `ftl` file, keep the structure in the template.

WDYT?

PR #1040, opened 2020-03-19
Fix footer link to respect localized pages

### LINKED ISSUE #1039: Link to documentation at the footer is not good for translated contents
### What needs to be fixed?

The link to documentation at the footer is linked to doc.rust-lang.org, then it redirects rust-lang.org/learn without respecting which page the user was visiting. 

image

It's not good for translated contents.For example, if the user was visiting Japanese content (i.e. rust-lang.org/ja), it should guide to rust-lang.org/ja/learn, not to rust-lang.org/learn.

### Page(s) Affected

All the pages that has footer. Technically,  [here](https://github.com/rust-lang/www.rust-lang.org/blob/master/templates/components/footer.hbs#L7)

### Suggested Improvement

change it to link to `{{baseurl}}/learn`

JohnTitor:
Fixes #1039

### DIFF
--- a/templates/components/footer.hbs
+++ b/templates/components/footer.hbs
@@ -4,7 +4,7 @@
       <div class="flex flex-column mw8 w-100 measure-wide-l pv2 pv5-m pv2-ns ph4-m ph4-l" id="get-help">
         <h4>{{fluent "footer-get-help"}}</h4>
         <ul>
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR!

PR #1043, opened 2020-03-21
Fix grammar

Smibu:
The PR https://github.com/rust-lang/www.rust-lang.org/pull/996 removed the word "and", but in that case, I think the verb needs to be changed too.


XAMPPRocky:
Good catch! Thank you for your PR, and congrats on your first contribution! 🎉

PR #1050, opened 2020-04-05
Switch embedded book to point to doc.rust-lang.org

XAMPPRocky:
Since we now publish the embedded book as a component we should link to it.


pietroalbini:
This looks good! Maybe let's get an opinion from the Embedded WG?

@rust-lang/wg-embedded-resources

pietroalbini:
Great then, merging this!

PR #1056, opened 2020-04-13
Add "Korean" Language

spaceranger-lightyear:
Hello, I'm one of Korean university student that interested in Rust and just started studying Rust.

I usually visit rust-lang.org but i recognize that there is no Korean.
So I translate and add "ko-KR" (Korean) as Korean.

Addtionally, there is a original link of policy/media-guide at ko-KR/media-guide.ftl

Thank you :)


Manishearth:
Hi,

translations are maintained on Pontoon, see https://github.com/rust-lang/www.rust-lang.org/blob/master/TRANSLATIONS.md

There already is a korean translation team, please talk to @sanxiyn and @lifthrasiir  about being added to it.

PR #1057, opened 2020-04-14
Update dependencies to remove some packages

JohnTitor:
It's minor clean-up, shouldn't break any behavior.
First commit is quite _large_ but it's just converting lockfile format, there's no actual change.


XAMPPRocky:
Thank you for your PR!

PR #1079, opened 2020-04-20
use <main> instead of a div with a id #main-content

vired:
use semantic html element


XAMPPRocky:
Thank you for your PR, and congrats on your first contribution! 🎉

mobop:
🎉

PR #1136, opened 2020-06-04
Make .highlight smaller to avoid splitting text

mjbshaw:
The headlines on the website are unnecessarily difficult to read due to the text being split in half by two different backgrounds with a hard border. This PR makes the highlight a little smaller and a little lower so it looks more like a thick underline. I find that this greatly helps the readability of the headlines.

Before:
before

After:
after

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -342,9 +342,9 @@ h3 {
 }
 
 .highlight {
-  height: 24px;
+  height: 12px;
   position: relative;
-  top: -24px;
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! Generally I'm in favour of this change, though I'll need to check if there's any places where this change breaks the existing design. If no one else has concerns and once I've done a cursory review of the impact of this change I'll merge this in (Which I'll get to probably sometime next week).

PR #1165, opened 2020-07-01
refine highlight's spacing and border-radius

XAMPPRocky:
Updates highlight spacing so it doesn't clash with the descent in some titles and updated the border radius to be visually consistent with the buttons.

### Production

Screenshot 2020-07-01 at 10 19 06

### Staging

Screenshot 2020-07-01 at 10 18 50

### Proposed Changes

Screenshot 2020-07-01 at 10 25 56

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -626,6 +626,13 @@ blockquote::before {
   font-weight: 600;
 }
 
+@media screen and (max-width: 769px) {
+  .highlight {
+    border-bottom-left-radius: 0;
+    border-top-left-radius: 0;
+  }
+}
+
 // Temporary HACK: switch away from this when we switch to using Tachyons for
 // the page titles -- at that time, we should make a simple component class that
 // just composes those classes.
@@ -634,11 +641,6 @@ blockquote::before {
   header {
     h1 {
       font-size: 4rem;
-    }
-
-    .highlight {
-      height: 20px;
-      top: -20px;
     }
 
     h2.subtitle {
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -344,11 +344,12 @@ h3 {
 .highlight {
   height: 12px;
   position: relative;
-  top: -6px;
+  top: 0;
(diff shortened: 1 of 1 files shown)


Manishearth:
The clashing of the highlight spacing was a deliberate design choice, fwiw. @ashleygwilliams do you have any thoughts on changing this?

r=me on the border change

XAMPPRocky:
I'm aware it's a deliberate choice. The motivation for this change is that the "highlight" works best on one line titles however the clash doesn't work in my opinion with multi-line titles and makes them harder to read, as it doesn't feel like it part of the title in the same way when it's a single line. This stands out particularly in the blog, I've pasted some screenshots of examples.

Screenshot 2020-07-04 at 17 22 01
Screenshot 2020-07-04 at 17 23 08


Screenshot 2020-07-04 at 17 17 57
Screenshot 2020-07-04 at 17 19 04

Screenshot 2020-07-04 at 17 19 48
Screenshot 2020-07-04 at 17 19 39

Manishearth:
(In the future, please include such motivation in the PR itself)

This seems fair, but it seems to largely be a concern for the mobile site where such wrapping is common. I wonder if we can media query it to only apply to the site on mobile? (in general I think the styles need some tweaks on mobile) I think it looks kinda nice on desktop, and I've seen a trend of websites using this kind of underline style elsewhere too.

I'd also still like to hear from Ashley.

XAMPPRocky:
> (In the future, please include such motivation in the PR itself)

This is not the PR that makes that change. #1136 had been open for over a month with no concerns raised and I had talked to other team members who were also in favour of the change. In the future I would appreciate leaving concerns on the appropriate PR before it's merged, instead of in follow up PRs.

Manishearth:
@XAMPPRocky okay, I wasn't aware of that PR, it must have slipped my notice. Nor was I aware other team members supported it. I do think we should do team pings for major styling changes, though (this PR is fine).

r+ on this and the blog PRs, then.

PR #1168, opened 2020-07-02
Mention installing Windows VS C++ build tools

ChrisDenton:
Adds a sentence about the C++ build tools being a prerequisite on Windows.

Alternatively it could link to the instructions in [the book](https://doc.rust-lang.org/book/ch01-01-installation.html#installing-rustup-on-windows) where we could (potentially) give more detailed install instructions. However, I think it's useful to new users to directly present everything they need to get started in Rust.

EDIT: Updated the language used to directly refer to following the prompts in the rustup installer.

### DIFF
--- a/locales/en-US/tools.ftl
+++ b/locales/en-US/tools.ftl
@@ -154,7 +154,7 @@ install-other-methods-link = Learn more
 ## components/tools/rustup.hbs
 
 tools-rustup-unixy = It looks like you’re running macOS, Linux, or another Unix-like OS. To download Rustup and install Rust, run the following in your terminal, then follow the on-screen instructions. See <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">"Other Installation Methods"</a> if you are on Windows.
-tools-rustup-windows = It looks like you’re running Windows. To start using Rust, download the installer, then run the program and follow the onscreen instructions. See <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">"Other Installation Methods"</a> if you are not on Windows.
+tools-rustup-windows-2 = It looks like you’re running Windows. To start using Rust, download the installer, then run the program and follow the onscreen instructions. You may need to install the <a href="https://visualstudio.microsoft.com/visual-cpp-build-tools/">Visual Studio C++ Build tools</a> when prompted to do so. If you are not on Windows see <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">"Other Installation Methods"</a>.
 tools-rustup-wsl-heading = Windows Subsystem for Linux
 tools-rustup-wsl = If you’re a Windows Subsystem for Linux user run the following in your terminal, then follow the on-screen instructions to install Rust.
 tools-rustup-unknown = Rust runs on Windows, Linux, macOS, FreeBSD and NetBSD. If you are on one of these platforms and are seeing this then please report an issue with the following values:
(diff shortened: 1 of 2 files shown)


Fasani:
I think this is a nice addition, I had to do this yesterday and I would go one step further and outline what needs to be installed.

Eg:
From "Visual Studio Build Tools 2019" I had to select these 2 options:
- "MSVC v142 - VS 2019 C++ x64/x86 build tools"
- "Windows 10 SDK (10.0.18362.0)"

But I am unsure if this is true for most people?

ChrisDenton:
My main motivation for this PR is to warn of the prerequisite before users start the installation process so that it doesn't come as a surprise. I'm unsure if we need to provide full install instructions here as rustup itself does that. But maybe we do.

The rustup installation instructions currently say the following:

> The easiest way to acquire the build tools is by installing Microsoft
> Visual C++ Build Tools 2019 which provides just the Visual C++ build
> tools:
>
>     https://visualstudio.microsoft.com/visual-cpp-build-tools/
>
> Please ensure the Windows 10 SDK and the English language pack components
> are included when installing the Visual C++ Build Tools.

This recommends installing the C++ Build Tools workload and adding two components on top of the default selection.

Fasani:
There is a good chance I overlooked it.... I spent around 30 minutes trying to figure this out on stackoverflow/google.

On my Mac it took me around 10 seconds :-)

ChrisDenton:
Oh for sure, you're not the only one. The installation story on Windows could be better or at least more automated. I doubt it'll get down to 10 seconds though as Rust does require platform tools (e.g. the linker and import libraries) that aren't available by default and can't be directly bundled (due to licensing, IIRC).

locales/en-US/tools.ftl
```diff
@@ -154,7 +154,7 @@ install-other-methods-link = Learn more
 ## components/tools/rustup.hbs
 
 tools-rustup-unixy = It looks like you’re running macOS, Linux, or another Unix-like OS. To download Rustup and install Rust, run the following in your terminal, then follow the on-screen instructions. See <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">"Other Installation Methods"</a> if you are on Windows.
-tools-rustup-windows = It looks like you’re running Windows. To start using Rust, download the installer, then run the program and follow the onscreen instructions. See <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">"Other Installation Methods"</a> if you are not on Windows.
+tools-rustup-windows = It looks like you’re running Windows. To start using Rust, download the installer, then run the program and follow the onscreen instructions. You may need to install the <a href="https://visualstudio.microsoft.com/visual-cpp-build-tools/">Visual Studio C++ Build tools</a> when prompted to do so. If you are not on Windows see <a href="https://forge.rust-lang.org/infra/other-installation-methods.html">"Other Installation Methods"</a>.
```
XAMPPRocky, line 157: Would you be able to change the key here and in the html to `tools-rustup-windows-2`? This will help by letting our translators know that this content needs to be updated.

Fasani: @XAMPPRocky just for my own interest, how does changing that key allow them to know? Cheers

XAMPPRocky: @Fasani Sure! We use a localisation service called pontoon <https://pontoon.rust-lang.org> and changing the key will create a new untranslated key for every other locale. So pontoon will show them they haven’t translated the new text.

ChrisDenton: Ok, I think that's fixed now?

XAMPPRocky:
Thank you again for your PR!

PR #1169, opened 2020-07-02
Fix #1137 - Changing rustup to Rust in the sentence 'Is Rust up to date?'

### LINKED ISSUE #1137: Getting started page says "Is rustup up to date?" not "Is Rust up to date?"
### What needs to be fixed?

The getting started page has a header that says "Is rustup up to date?" where I would expect it to say "Is Rust up to date?".

I'm not sure if this is intentional, but I figured I should point it out because it seemed confusing to me.

### Page(s) Affected

[Getting started](https://www.rust-lang.org/learn/get-started)

### Suggested Improvement

Replace the header "Is rustup up to date?" with "Is Rust up to date?".

Fasani:
Fixes:
https://github.com/rust-lang/www.rust-lang.org/issues/1137

The discussion on 1137 seems to suggest the sentence should be `Is Rust up to date?` as a native English speaker I also believe this to be the correct change.

I do not speak all of the other languages but the change to me would seem ok on all language versions.


XAMPPRocky:
Thank you for your PR!

PR #1170, opened 2020-07-03
Move Windows dependency note to the top of Notes

notriddle:
https://users.rust-lang.org/t/no-entry-on-rust-lang-web-site-about-msvs-builds-tools-and-windows-sdk/45263


XAMPPRocky:
Thank you for your PR!

PR #1185, opened 2020-07-11
Updates favicons to support dark mode

### LINKED ISSUE #1157: Adaptive favicon (light vs. dark mode)
### Summary

The current favicon is not quite visible when the dark system theme is used (I'm currently using macOS's dark mode).

Screenshot 2020-06-25 at 20 23 56

The proposal is to have an adaptive favicon, one that visibly changes the favicon based on the system's current theme using the `prefers-color-scheme` media query and a bit of JavaScript.

Light mode:

Screenshot 2020-06-25 at 20 17 43

Dark mode:

Screenshot 2020-06-25 at 20 17 23

### Motivation

To easily identify the browser tabs that are opened to Rust websites.

### Drawbacks

People might not like the light Rust logo in dark mode.

This also uses a CSS media query which is not supported on all versions of Firefox for Android, Samsung Internet, and IE. It's also not supported on the versions prior to:

* Chrome 76
* Edge 76
* Firefox 67
* Safari 12.1

See more [here](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).

As a workaround, we can have it fallback to the default favicon that we see now (see `Rationale and alternatives`)

An alternative solution is to create a favicon that explicitly has a light background inside the Rust logo. Similar to how the favicon looks in light mode, the dark mode will still show dark Rust logo with a light background.

This requires no CSS media query feature support and no extra JavaScript.

### Maintenance

I've already created a fix for this issue that involves using the CSS media query and JavaScript.

### Unresolved Questions

N/A

sl4m:
This fixes #1157 using only `prefers-color-scheme` CSS media query. This was tested on the latest Firefox, Chrome, and Microsoft Edge.

I modified the existing favicon.svg to include CSS that will detect the system color scheme and appropriately apply the CSS. It seems the order in which the icons are referenced in the template matters, so I've re-ordered them in a way that worked during testing. For browsers that do not support SVG, it falls back on the modified png favicons which now have a white background and works on either system color scheme. Also, the reference to favicon.ico has been removed in the template as it will take precedence over all other available favicons.

I'm happy to make these changes for all the other repos that use the Rust logo favicon. I believe it is:

* blog
* thanks
* ???

Chrome / Microsoft Edge:
Screenshot 2020-07-10 at 18 34 11
Screenshot 2020-07-10 at 18 28 35

Firefox:
Screenshot 2020-07-10 at 18 28 02
Screenshot 2020-07-10 at 18 29 36

PNG fallback:
Screenshot 2020-07-10 at 17 46 58
Screenshot 2020-07-10 at 17 47 22


XAMPPRocky:
Thank you for your PR! The PNG fallback had me thinking, do you think it would be possible to have the dark mode version have just a white border and white fill in through CSS? I was thinking something along these lines though I would probably give the favicon a thicker outline than in the picture.

rust-dark

sl4m:
@XAMPPRocky I spent some time trying to figure out how to manipulate the SVG to support what you've requested. I believe the SVG needs additional elements to support a border and fill. Unfortunately, my SVG knowledge stops here. If someone else can create the SVG with the additional elements, I can apply the CSS appropriately.

XAMPPRocky:
@sl4m I suspected that would be case. Either way I think this fine to merge Thank you for your PR and congrats on your first contribution!

sl4m:
@XAMPPRocky What are your thoughts about propagating these changes to other repos that are associated with rust-lang.org's subdomains? This is not a complete list, but I'm thinking:

* blog.rust-lang.org
* thanks

XAMPPRocky:
I think it would be fine to make a PR for each of those.

PR #1188, opened 2020-07-12
Partial revert change and improve responsive layout

Fasani:
@XAMPPRocky Hey, I noticed my earlier commit (11 days ago) had side effects on larger screens, buttons were no longer visible, I have reverted it and wrapped the buttons in an additional div/class (see additional comment).

Demo:
https://www.loom.com/share/cf6966b43e844258b04459df22ff5b81

### DIFF
--- a/templates/community/index.hbs
+++ b/templates/community/index.hbs
@@ -82,6 +82,29 @@
         </ul>
       </div>
     </div>
+    <div class="dn db-l">
+      <div class="flex flex-column flex-row-l">
+        <div class="w-33-l mh3-l pt0 flex flex-column justify-start" id="users-forum">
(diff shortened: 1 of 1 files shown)


templates/community/index.hbs
```diff
@@ -82,6 +82,29 @@
         </ul>
       </div>
     </div>
+    <div class="dn db-l">
```
Fasani, line 85: The original code prior to 11 days ago had `.dn-m` on line 86 the problem is that that the flex class overrides that `.dn-m` for specificity reasons. So it needs to sit higher in its own container.

So the rule on line 85 now says `display: none` and `display:block on large screen`.

XAMPPRocky:
Thank you for your PR!

PR #1238, opened 2020-08-19
Add uninstall information

### LINKED ISSUE #1213: Add instructions to uninstall rust
### Summary

We should add helpful uninstall instructions, pointing out that this might be useful from time to time, e.g to reclaim space or when facing trouble with a broken installation.

I found `rustup self uninstall` works well ([https://users.rust-lang.org/t/how-to-uninstall-rust/149/7](https://users.rust-lang.org/t/how-to-uninstall-rust/149/7?u=asfaltboy))

To be clear, this easy uninstall flow is more than most programming language environments offer, and I love it!

### Motivation

I remember installing rust on my Mac using a bash script (something like `curl -L https://static.rust-lang.org/rustup.sh | sudo sh`), but I could not figure out how to uninstall it (had to, when running close to out of disk space).

When trying to read through the docs and searching forums, I found a few mentions of `uninstall.sh` script or using a package manager (my HomeBrew showed that rust wasn't installed through it). It took me probably 15 minutes to find the `rustup` solution above.

I also considered trashing the files in `~/.cargo` manually, but got worried about finding any shared package paths.

### Drawbacks

Perhaps the requirement to maintain instructions to uninstall rust in this place, duplicating some other doc (e.g under `rustup` doc space), can be considered a drawback. Though we could link to this doc saying "Up to date information about the `uninstall` command can be found _here_".

We could also make a separate section for these instructions, or place them under "getting started". For me, though, it made the most sense to look under the installation section or use the search button.

### Maintenance

The `rustup` tool maintainers could be responsible to maintain this, along with other rustup commands on the website?

### Unresolved Questions

* Will this command work on all operating systems? Or, should we provide per-OS instructions (e.g uninstaller on windows, or package manager uninstaller)
* Should we provide a warning to the users about the dangers of uninstalling?

camelid:
Fixes #1213.

r? @XAMPPRocky

### DIFF
--- a/locales/en-US/tools.ftl
+++ b/locales/en-US/tools.ftl
@@ -99,7 +99,7 @@ install-notes-rustup-description =
 install-notes-uninstall-heading = Uninstall Rust
 install-notes-uninstall-description =
         <p>
-          If you at any point you would like to uninstall Rust, you can run
+          If at any point you would like to uninstall Rust, you can run
           <code>rustup self uninstall</code>.
           We'll miss you though!
         </p>
--- a/locales/en-US/tools.ftl
+++ b/locales/en-US/tools.ftl
@@ -96,6 +96,14 @@ install-notes-rustup-description =
           <code>rustup</code> documentation</a>.
         </p>
 
+install-notes-uninstall-heading = Uninstall Rust
+install-notes-uninstall-description =
+        <p>
+          If you at any point you would like to uninstall Rust, you can run
+          <code>rustup self uninstall</code>.
+          We'll miss you though!
+        </p>
+
 install-notes-path-heading = Configuring the <code>PATH</code> environment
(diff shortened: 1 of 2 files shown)


camelid:
Here's what it looks like:

image

camelid:
Do you think it should be a subsection under "Toolchain management with `rustup`"?

asfaltboy:
I personally prefer this as a heading to make it easier to see. Kudos for the initiative 👍 !

XAMPPRocky:
Thank you for your PR! I think the subsection is fine as is though I would prefer it at the end of the section, below "configuring your path". I usually expect un-installation instructions to be at the end of a page.

XAMPPRocky:
Thank you again for your PR!

PR #1239, opened 2020-08-20
Update nav.hbs

pseudo-su:
I don't think the difference between the padding on either side here is intentional. It makes the logo not quite centered.

Before:

before

After:

after


XAMPPRocky:
Nice catch! Thank you for your PR, and congrats on your first contribution! 🎉

PR #1259, opened 2020-09-02
Remove horizontal rule from "Generating a new project" on "Getting started" page

camelid:
It looks out of place and a bit weird in this spot to me.

@rustbot modify labels: A-Design


camelid:
# Before

image

# After

image

Manishearth:
It seems fine to me, it's being separated into two logical steps. Otherwise it kinda looks like a list of `cargo` things with the `cargo run` not having any text with it, which is confusing

camelid:
Okay, would it be okay if I changed the color of the rule to white to match the text? I think the gray against the red just looks weird to me.

Manishearth:
Yes

senekor:
superseded by https://github.com/rust-lang/www.rust-lang.org/pull/1966

PR #1262, opened 2020-09-02
Fix syntax highlighting

camelid:
Add the `plaintext` language class to some code blocks that were being
highlighted incorrectly. The `cargo run` one was being highlighted as
Ruby and the `ferris_says` one was being highlighted as Bash!

Also added a missing `p` tag around an inline code block that was on its
own line.

### DIFF
--- a/templates/learn/get-started.hbs
+++ b/templates/learn/get-started.hbs
@@ -46,13 +46,13 @@
     </header>
     {{#fluent "learn-generating-steps"}}
     {{#fluentparam "tree"}}
-    <pre><code>hello-rust
+    <pre><code class="plaintext">hello-rust
 |- Cargo.toml
 |- src
   |- main.rs</code></pre>
     {{/fluentparam}}
     {{#fluentparam "output"}}
-    <pre><code>$ cargo run
+    <pre><code class="plaintext">$ cargo run
    Compiling hello-rust v0.1.0 (/Users/ag_dubs/rust/hello-rust)
     Finished dev [unoptimized + debuginfo] target(s) in 1.34s
      Running `target/debug/hello-rust`
@@ -100,7 +100,7 @@ fn main() {
     </code></pre>
     {{/fluentparam}}
     {{#fluentparam "output"}}
-    <pre><code>----------------------------
+    <pre><code class="plaintext">----------------------------
 | Hello fellow Rustaceans! |
 ----------------------------
               \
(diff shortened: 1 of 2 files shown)


camelid:
I have to admit that the `ferris_says` one was kind of cute with bash syntax highlighting:

image

camelid:
# Before

image

image

# After

image

image

templates/learn/get-started.hbs
```diff
@@ -46,13 +46,13 @@
...
     {{#fluentparam "tree"}}
-    <pre><code>hello-rust
+    <pre><code class="plaintext">hello-rust
 |- Cargo.toml
 |- src
   |- main.rs</code></pre>
     {{/fluentparam}}
     {{#fluentparam "output"}}
-    <pre><code>$ cargo run
+    <pre><code class="plaintext">$ cargo run
```
camelid, line 55: I tried using the `console` and `shell` language classes, which are supposedly [supported](https://github.com/highlightjs/highlight.js/blob/fc33aea76518928f791589f01371908949c332aa/SUPPORTED_LANGUAGES.md) by Highlight.js, but I guess our build doesn't have them? It would be nice to use that instead so that it's not just a wall of text.

camelid: @XAMPPRocky Do you think you could add Shell to our custom build?

XAMPPRocky: Seems sensible to me. Changing the custom build is easy and I think would make sense as part of the PR. You can download a new one here, just uncheck all the unwanted languages, and replace the static files. https://highlightjs.org/download/

camelid: Hmm, do we still need YAML support?

XAMPPRocky: Probably not.

camelid: Okay then, I think I'll do:

* shell session
* plaintext
* rust
* bash
* toml
* markdown

camelid: Only the `$` was syntax highlighted 😄:

image

camelid:
@XAMPPRocky Note that the latest version of hljs is 10 and we were previously on 9. I wrote a very lengthy commit message in a8e1d06 about hljs, but it would probably be good to double-check their upgrade guide here: https://github.com/highlightjs/highlight.js/blob/master/VERSION_10_UPGRADE.md

XAMPPRocky:
Changes seem sensible to me. I'm pretty sure IE 11 isn't in our compatibility list, so I don't think that's a concern. Thank you again for your PR!

PR #1266, opened 2020-09-11
Fix various typos found using `codespell -wi3`

Calinou:


XAMPPRocky:
Thank you for your PR!

PR #1291, opened 2020-10-01
policies/privacy: Remove `#privacy-notice` ID from section element

Turbo87:
It turns out that certain adblockers (e.g. uBlock Origin) include the "EasyList Cookies" blocklist, which prevents elements with an `#privacy-notice` ID from rendering, which resulted in us showing a blank page. Since this ID does not seem to be used for anything here this commit just removes it to prevent the adblocker issue.

/cc @pietroalbini

### DIFF
--- a/templates/policies/privacy.hbs
+++ b/templates/policies/privacy.hbs
@@ -18,7 +18,7 @@
 </section>
 {{/if}}
 
-<section id="privacy-notice" class="purple">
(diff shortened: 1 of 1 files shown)


pietroalbini (approved):

pietroalbini:
That block list is... far from accurate :)

Issue #1292, opened 2020-10-01
Some HTML tags not rendered in Spanish /sponsors


Zheoni:
### Page(s) Affected

https://www.rust-lang.org/es/sponsors

### What needs to be fixed?

In the Spanish (es) website,  in the sponsors page, there are html tags not rendered in several paragraphs. These errors are not found in the English (en-US) website.
The problems are in the GitHub, Integer 32, and ARM paragraphs. \<em\> and \<code\> tags.

dzhibas:
There is 2 solutions to this one, and for both i will open PR. please choose one and decline another

1. disable sponsors description template variable escaping
2. remove html tags from spanish sponsors translation file

dzhibas:
PRs opened for both solutions. Pick one & close another one

https://github.com/rust-lang/www.rust-lang.org/pull/1295
https://github.com/rust-lang/www.rust-lang.org/pull/1294

senekor:
The sponsors page now redirects to the website of the Rust foundation.

PR #1294, opened 2020-10-02
#1292 sponsors description might have html no escape

dzhibas:
disables sponsors description escaping, to properly render spanish html tags in some sponsors

PR #1295, opened 2020-10-02
#1292 sponsors description might have remove html

dzhibas:
Remove html tags from spanish translation file as output is escaped for description and then its not properly rendered


dzhibas:
choose one of 2 solutions:
https://github.com/rust-lang/www.rust-lang.org/pull/1295
https://github.com/rust-lang/www.rust-lang.org/pull/1294

the other one close

dzhibas:
declining this one as other solution was taken

PR #1302, opened 2020-10-03
Removing debugging line

dzhibas:
Removing debugging println which was printing foo on each /sponsors request


dzhibas:
@Manishearth would you be able to add label `hacktoberfest-accepted` to this PR please

PR #1304, opened 2020-10-05
Fixes #767 - restructuring redirects

### LINKED ISSUE #767: Change redirect implementation
# Technical Feature Request
## Summary
Redirection is currently a simple match table from filename to filename. It would be better to configure them through a configuration file that goes (regex) -> (replacement).

This file could also act as a history of past redirects.

## Motivation
The current redirect systems is a set of functions together with a match table for the `<dest>` parameter. That has multiple drawbacks, including the necessity to have multiple redirect functions.

dzhibas:
Fixes #767

- moving redirects into yaml file
- redirects support regex matching and named&non-named group captures. groups can be passed to redirect url
- moved all  previous redirects into yaml
- removed previous routes for redirects which is moved
- redirect logic is moved into 404 catcher before throwing 404 we check if we  have a redirect match else throw 404 with custom template

### DIFF
--- a/src/data/redirects.yml
+++ b/src/data/redirects.yml
@@ -1,0 +1,82 @@
+---
+
+# Website Redirects setup
+# available keys: from, to, status[optional]
+#
+#   from - is a regex with names & non-named group captures
+#   to - is redirecting to with references to groups as $1 $2 or by name ifi it was captured like (?P<name>.*)
+#   status[optional] - is redirect response status, if not specified default value 302 Found will be used
+#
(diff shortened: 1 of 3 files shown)


dzhibas:
if you like this PR guys and thinking on taking this approach a go, please tag this PR with label `hacktoberfest-accepted` thanks

dzhibas:
closing this one as it was re-implemented different way in this PR https://github.com/rust-lang/www.rust-lang.org/pull/1529

Issue #1274, opened 2020-09-18
Wrong display for escaped characters in Zulip streams


LeSeulArtichaut:
### What needs to be fixed?
URL-escaped characters, e.g. `/`, are displayed as their escaped counterpart in Zulip stream names.

### Page(s) Affected
All `/governance` teams/WGs who have attached Zulip streams containing an escaped character. For example, see all WGs that are subteams of `T-compiler`.

### Suggested Improvement
The website should probably use the raw data from `rust-lang/team`. #1235 might be reworked to store both displayable strings and escaped strings to create the links.

camelid:
Can you give an example of one of the WGs under the t-compiler stream? I checked pretty much all of the WGs on /governance and none of them are under the t-compiler stream.

LeSeulArtichaut:
Here's an example:

![](https://user-images.githubusercontent.com/38361244/98582895-da37f700-22c3-11eb-8d7d-107ef55874ff.png)

Notice how the Zulip button reads `#t-compiler.2fwg-diagnostics` instead of `#t-compiler/wg-diagnostics`

camelid:
This is actually much more involved than I thought at first :/

@rustbot release-assignment

LeSeulArtichaut:
I guess this could be solved in `rust_team_data` by doing the URL conversion thing there, so that both the name and the escaped link are availible in `Team`.

PR #1344, opened 2020-11-13
Fix escaping for Zulip streams

jryans:
This moves the escaping for Zulip streams out of the data pipeline and over to a
Handlebars helper. This in turn allows the group team template to make use of
both the unescaped (for display) and escaped (for the link) variants.

image

Fixes https://github.com/rust-lang/www.rust-lang.org/issues/1274

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -34,6 +34,7 @@ mod sponsors;
 mod teams;
 
 use production::User;
+use teams::encode_zulip_stream;
 
 use std::collections::hash_map::DefaultHasher;
 use std::env;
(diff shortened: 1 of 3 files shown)


pietroalbini (approved):

pietroalbini:
Thanks!

jryans:
Looks like the fix is live on staging: e.g. https://www-staging.rust-lang.org/governance/teams/compiler, so it should fix the production site whenever the next deploy happens.

pietroalbini:
A deploy is happening: https://github.com/rust-lang/www.rust-lang.org/pull/1348

PR #1370, opened 2020-12-05
More direct link to translated Rust books (french version in my case)

luidnel-maignan:
Follow up on the issue https://github.com/rust-lang/www.rust-lang.org/issues/1365


locales/fr/learn.ftl
```diff
@@ -3,7 +3,9 @@
 learn-page-title = Apprendre Rust
 learn-begin = Commencer avec Rust
 learn-book = Affectueusement surnommé le "Book", <cite>The Rust Programming Language</cite> vous donnera une vue d'ensemble du langage depuis ses premiers principes. Vous construirez quelques projets au long du chemin, et à la fin, vous aurez une solide compréhension du langage.
-learn-book-button = Lire le Book !
+learn-book-button = Lire le Book officiel !
```
Manishearth, line 6: I would still just call it `le Book` here

Manishearth:
Can you post a screenshot of what this looks like

luidnel-maignan:
image

luidnel-maignan:
To be frank, I would be to obtain the following, but I haven't tried yet.

image

Manishearth:
cc @pietroalbini do you have opinions on how this should be presented?

skade:
With no judgement about the quality of the translation, we followed the general rule of not linking to external sources with changing maintainership when building the new website. That was one of the major banes of the old website: most links were bad or wrong, as the material on the other side either vanished or became stale. It comes with a huge burden to keep those resources checked.

That being said: I think especially things as notable as full book translations should find a way it our material. Here, I would prefer if it _also_ made its way into our English material somehow and if there were some way to list translated book versions in a central place.

Manishearth:
I do think the translated link should be less prominent, but not sure how to do that. As a bare minimum those buttons should definitely be aligned

Manishearth:
> That being said: I think especially things as notable as full book translations should find a way it our material. Here, I would prefer if it _also_ made its way into our English material somehow and if there were some way to list translated book versions in a central place.

I do kinda agree with this: @steveklabnik thoughts on if this should be linked to from the book somehow?

locales/fr/learn.ftl
luidnel-maignan: My feeling was that a nice warning was needed before a change of language. But I can remove "officiel" if you want.

pietroalbini:
Another possible representation (hacked with CSS):

luidnel-maignan:
This looks good to me and seems to meet all current requirements ! Thank you @pietroalbini !
So what is the next step ?

Manishearth:
@luidnel-maignan can you implement it to look like what Pietro suggested?

luidnel-maignan:
I can try. @pietroalbini, can you share your CSS hack ?

pietroalbini:
@luidnel-maignan my CSS ***hack*** was to add `margin-bottom: 30px` to the other two columns to make it kinda look good, which I don't think we should ship in production :upside_down_face:

I'm not a big CSS expert, but I'm not sure whether that layout is achievable with flexbox. We might have to switch to CSS Grids for it.

luidnel-maignan:
Thank you @pietroalbini, I made a similar thing by adding a div for the button(s) in each of the three boxes, and ask them to have a convenient min-height. You can try the result on http://46.226.108.172:8002/fr/learn

steveklabnik:
https://github.com/rust-lang/book#translations is the sort of "official" link for translations, it includes relevant info and a link to all transaltions.

luidnel-maignan:
What is the next step for this update ? Is there anything to improve or should we just try to merge like this and see the feedbacks ?

Manishearth:
@pietroalbini are you fine merging as is?

src/styles/app.scss
```diff
@@ -180,6 +198,10 @@ section {
   }
 }
 
+#learn-begin {
+  padding: 30px 0 0 0;
```
pietroalbini, line 202: This approach is really fragile, and I'm sure it will break eventually. I did use it to prototype the change, yes, but in the actual change we land I'd prefer a more robust approach.

luidnel-maignan: I agree, I think it's ok to just remove these lines.

luidnel-maignan: I removed them and also reduce the min-height of buttons from 100px to 70px.
This works really well, the result can be viewed at http://46.226.108.172:8002/fr/learn as usual.

pietroalbini (approved):

luidnel-maignan:
Cool, thank you all to have taken the time and the care for this little thing !

PR #1419, opened 2021-01-25
Sort subteams by website.weight.

m-ou-se:
The `website.weight` from the `teams.json` data was not used for subteams. This fixes that.

Right now there's not any subteam with a weight, but there might be soon when we make some changes to the libs (sub)team(s).


KodrAus:
We could probably add some weights to subteams in the `test_page_data` test for good measure.

pietroalbini:
Yeah, adding a test for this would be great!

m-ou-se:
Done!

pietroalbini (approved):

PR #1502, opened 2021-04-16
feat(policies): update privacy policy to foundation page

ashleygwilliams:
It is regrettable that this change disrupts the translations, but this is a necessary update. Hopefully we can get translations working on the Foundation website, but we'll need to wait til there's slightly less churn on the policies.


Mark-Simulacrum (approved):
The change looks good to me, but I would like someone more familiar with the website infrastructure (e.g., translation layer) to confirm this is good before we merge.

Manishearth (approved):

pietroalbini:
Could you also add server-side redirects to make sure no links are broken? Redirects are implemented in [`src/redirects.rs`](https://github.com/rust-lang/www.rust-lang.org/blob/master/src/redirect.rs).

pietroalbini (approved):

pietroalbini (approved):
Thanks!

Issue #1504, opened 2021-04-16
Missing redirect for the privacy policy


pietroalbini:
We need to make sure redirects from the old privacy notice page to the foundation's privacy policy (introduced in #1502) correctly work. Right now they don't work because [the route only matches a single segment](https://github.com/rust-lang/www.rust-lang.org/blob/7db534b1c745a94a087f56ffa27be8cf00489f72/src/main.rs#L276-L279).

The fix for this should be simple (change the route to allow multiple segments), but we should be careful to make sure doing that doesn't change the routing priority.

AliSajid:
This can be closed as the pull request worked!

PR #1529, opened 2021-05-15
Refactor redirects to be more consistent

### LINKED ISSUE #767: Change redirect implementation

pietroalbini:
This commit does a complete refactoring of how the website handles redirects, fixing multiple problems with them:

* Redirects didn't handle localizations at all, always redirecting to the English website even when the proper localization exists.
* Redirects didn't work for multiple path segments, only supporting a single segment instead. For example no redirects on `/foo/bar` could have been created.
* Redirects didn't work if another route processed the URL and returned 404. For example this prevented a redirect for `/policies/privacy` as the request was already processed by `/policies/:name`.

The main changes in this commit are:

* Redirects are now handled in the 404 handler instead of dedicated redirect handlers with a lower priority: this allows redirects to be performed even when another handler processes the requests and explicitly returns a 404 status code.
* Redirects now support paths with an arbitrary amount of segments.
* Redirects now properly handle localizations: they work inside existing locales, they convert pre-2018 locales to current locales and only redirect to the English version when a pre-2018 locale does not exist anymore on the current codebase.

Fixes #767 #1504
r? @Manishearth

### DIFF
--- a/src/i18n.rs
+++ b/src/i18n.rs
@@ -90,7 +90,7 @@ pub const EXPLICIT_LOCALE_INFO: &[LocaleInfo] = &[
 ];
 
 lazy_static! {
-    static ref SUPPORTED_LOCALES: HashSet<&'static str> =
+    pub static ref SUPPORTED_LOCALES: HashSet<&'static str> =
         EXPLICIT_LOCALE_INFO.iter().map(|x| x.lang).collect();
 }
 pub struct TeamHelper {
--- a/src/main.rs
+++ b/src/main.rs
@@ -273,38 +273,17 @@ fn load_users_data() -> Vec<Vec<User>> {
     users.chunks(3).map(|s| s.to_owned()).collect()
 }
 
-#[get("/<dest>", rank = 19)]
-fn redirect(dest: redirect::Destination) -> Redirect {
-    Redirect::permanent(dest.uri)
-}
-
-#[get("/pdfs/<dest>")]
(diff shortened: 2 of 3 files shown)


pietroalbini:
I tested the various code paths locally and they all seem to work!

Manishearth (approved):

dzhibas:
have you checked this redirect refactor? https://github.com/rust-lang/www.rust-lang.org/pull/1304 feels like mine was a bit cleaner with same hooking into 404 but then allows to maintain regex redirects with group replacements and stuff in yaml

i will close mine if there is no more plan of using it then

pietroalbini:
Hi @dzhibas, and I'm so sorry I didn't see your PR :(

The YAML in #1304 is indeed nice, but this PR also fixes redirects inside locales and redirects from old locales to new locales (for example `/it-IT/documentation.html` to `/it/learn`). To be fair that wasn't tracked in the issue you tried fixing, but addressing that requires to either make the regexes way more complex or an approach like this PR (with three different sets of redirects and the locale logic handled in code).

dzhibas:
ok, no worries ill close mine then

PR #1670, opened 2022-05-27
Use TLSv1.3 with curl if specified at all

pothos:
The curl option specified to use TLSv1.2 explicity while nowadays 1.3
is availalble and recommended.
Switch to specifying 1.3 instead of 1.2 for the command that downloads
the install script.

See https://github.com/rust-lang/book/pull/3130 and https://github.com/rust-lang/rustup/pull/2996

### DIFF
--- a/templates/components/tools/rustup.hbs
+++ b/templates/components/tools/rustup.hbs
@@ -1,7 +1,7 @@
 <div class="row">
   <div id="platform-instructions-unix" class="instructions db">
     <p>{{fluent "tools-rustup-unixy"}}</p>
-    <pre><code class="db w-100">curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh</code></pre>
+    <pre><code class="db w-100">curl --proto '=https' --tlsv1.3 -sSf https://sh.rustup.rs | sh</code></pre>
   </div>
   <div id="platform-instructions-win" class="instructions dn">
     <p>{{fluent "tools-rustup-windows-2"}}</p>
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Do you have links/pointers to changelogs for curl and the likelihood of this option being available in distros etc? I'd want us to avoid suggesting a command that will fail for users in practice due to an older curl locally, particularly when (AIUI) the security benefits in this case are likely to be fairly small.

cuviper:
For a quick sample -- RHEL7 has `curl --tlsv1.3`, but RHEL6 does not.

cuviper:
https://daniel.haxx.se/blog/2018/03/27/play-tls-1-3-with-curl/
> [curl and libcurl](https://curl.haxx.se/) has supported an explicit option to select TLS 1.3 since curl 7.52.0 (December 2016) and assuming you build curl to use a TLS library with support, you’ve been able to use TLS 1.3 with curl since at least then. The support has gradually been expanded to cover more and more libraries since then.

RHEL7 only has curl 7.29.0, but the `--tlsv1.3` option was backported.

cuviper:
Also, curl/curl#2994 updated the help text to say that these options are all "or greater", implying that was already the case, so `--tlsv1.2` would still allow a TLSv1.3 connection if possible.

pothos:
Makes sense to leave it at 1.2 for longer then. Primary motivation was to have the usage of the command aligned in the website, book, and the rustup website/docs, and a PR for this got merged for the book.

For security concerns I assume that the server uses the ciphers for 1.2 that are marked secure in https://en.wikipedia.org/wiki/Transport_Layer_Security#Cipher - Not sure about the details of possible TLS 1.2 MITM attacks but I think the TLS cipher matters only if we assume there is a possible attack where the correct server gets connected and the cert gets validated but later in the connection the attacker would inject something and send wrong data (Does this even exist?). In other cases like a malicious server/certificate being accepted in the connection then the cipher selection doesn't matter much. So I would say as long as HTTPS is used it's not really important here to enforce TLS 1.3.

PR #1795, opened 2023-04-07
migrate to rocket 0.5.0-rc.3

remlse:


Dockerfile
```diff
@@ -51,7 +51,7 @@ COPY static /app/static/
 COPY src/styles /app/src/styles/
 
 ENV ROCKET_PORT 80
-ENV ROCKET_ENV prod
+ENV ROCKET_PROFILE release
```
remlse, line 54: From the [changelog](https://github.com/SergioBenitez/Rocket/blob/master/CHANGELOG.md#silent-changes):

> ROCKET_ENV is now ROCKET_PROFILE. A warning is emitted a launch time if the former is set.

Rocket.toml
```diff
@@ -1,3 +1,3 @@
-[development]
+[debug]
```
remlse, line 1: From the [changelog](https://github.com/SergioBenitez/Rocket/blob/master/CHANGELOG.md#silent-changes):

> The default profile for debug builds is now `debug`, not `dev`.

src/cache.rs
```diff
@@ -14,14 +15,22 @@ lazy_static! {
...
-        update_cache(generator)
+        task::spawn_blocking(move || {
+            update_cache::<T>(generator)
+                // stringify the error to make it Send
+                .map_err(|e| e.to_string())
+        })
+        .await
+        .map_err(Box::new)?
+        // put the previously stringified error back in a box
+        .map_err(|e| e.as_str().into())
```
remlse, line 33: I tried to make the caching more idiomatic for async, but that got very hairy very quickly... after many wasted hours, I gave up. Blocking in the foreground on a request should still only happen once for each cache entry when it's empty the first time.

src/caching.rs
```diff
@@ -1,40 +1,24 @@
...
-            directives,
+impl CachedNamedFile {
+    pub fn max_age(file: NamedFile, max_age: u32) -> Self {
+        Self {
+            file,
+            header: Header {
+                name: Uncased {
+                    string: "CacheControl".into(),
+                },
+                value: format!("max-age={max_age}").into(),
```
remlse, line 20: The `CacheDirective` was removed from Rocket. (A breaking change not found in the changelog :disappointed: ) Since this module isn't really used as generically as it was designed, I implemented a simpler, more manual version which is sufficient for how it's used.

I haven't tested it though, I might need to make sure this header actually works as intended.

src/category.rs
```diff
@@ -1,12 +1,11 @@
 use std::path::PathBuf;
 
-use rocket::http::RawStr;
 use rocket::request::FromParam;
 
 fn is_category(name: &str) -> bool {
     let mut path = PathBuf::from("templates");
     path.push(name);
-    path.push("index.hbs");
+    path.push("index.html.hbs");
```
remlse, line 8: `rocket_dyn_templates` (previously part of `rocket_contrib`) [sets content type based on the second file extension](https://docs.rs/rocket_dyn_templates/0.1.0-rc.3/rocket_dyn_templates/#content-type). So I renamed all the templates, otherwise the html would be sent as `text/plain`.

src/category.rs
```diff
@@ -27,17 +26,12 @@ impl Category {
...
-            Ok(url) => {
-                if is_category(&url) {
-                    Ok(Category { name: url })
-                } else {
-                    Err(format!("No category called <{}>", url))
-                }
-            }
-            Err(e) => Err(format!("URL illegal in utf-8 ({})", e)),
+    fn from_param(param: &'r str) -> Result<Self, Self::Error> {
+        let url = param.to_string();
```
remlse, line 30: from the changelog:

> `&RawStr` no longer implements `FromParam`: use `&str` instead.

src/i18n.rs
```diff
@@ -117,60 +113,59 @@ impl HelperDef for TeamHelper {
...
+        let Some(id) = id.relative_path() else {
             return Err(RenderError::new(
                 "{{team-text}} takes only identifier parameters",
             ));
         };
         let team = rcx
-            .evaluate_in_block_context(name)?
-            .ok_or_else(|| RenderError::new(format!("Cannot find team {}", name)))?;
+            .evaluate(context, name)
+            .map_err(|e| RenderError::from_error(&format!("Cannot find team {}", name), e))?;
```
remlse, line 142: This part is basically guesswork on my part. I have zero experience with handlebars, so these helper idioms are unknown to me. `evaluate_in_block_context` doesn't exist anymore. I found no hints about this in the changelog of handlebars and `evaluate` isn't even documented in the current version :disappointed: .

I'm guessing these teams are related to the governance pages though, and they seem to work fine (after some debugging).

Manishearth:
Dockerfile needs fixing for rust-toolchain it seems

Manishearth (approved):
I don't see any issues with this.

Should get autodeployed to https://www-staging.rust-lang.org/ around 10 minutes after merging, we should play with that before we deploy properly

remlse:
image

@Manishearth oops... can you check the logs?

Manishearth:
@pietroalbini I no longer have access to the website/staging heroku, can you check?

pietroalbini:
@Manishearth granted back access to Heroku!

remlse:
Hmm. As far as I can tell, the handling of the `ROCKET_PORT` environment variable did not change. Nothing about it in the [changelog](https://github.com/SergioBenitez/Rocket/blob/master/CHANGELOG.md), [upgrade guide](https://rocket.rs/v0.5-rc/guide/upgrading/) or [configuration docs](https://rocket.rs/v0.5-rc/guide/configuration/).

The dockerfile specifies
```
ENV ROCKET_PORT 80
```
but Heroku's Procfile overrides that...? (I have no experience with Heroku)
```
web: ROCKET_PORT=$PORT ROCKET_PROFILE=release ./target/release/www-rust-lang-org
```
Not sure which is correct. Either way, I'm confused as to how that was impacted by the rocket migration.

Manishearth:
I don't think Heroku looks at the dockerfile at all, yeah, that's just there for local dev and CI. I think?

Manishearth:
http://www-staging.rust-lang.org/ seems like that worked!

Manishearth:
We should update the Rocket.toml with that instead of using an env var but I'll do that later

Manishearth:
Also we should definitely click around the website and ensure basically all of it works before deploying to prod.

Manishearth:
One thing from the logs:

```
2023-04-10T17:50:22.913479+00:00 app[web.1]:    >> Parameter guard `category: Category` is forwarding: "No category called <zh-TW>".

2023-04-10T17:50:23.046223+00:00 heroku[router]: at=info method=GET path="/static/styles/vendor_10880690442070639967.css" host=www-staging.rust-lang.org request_id=d7ad0c83-cb43-414f-81f6-33e4bafd572f fwd="24.7.113.20" dyno=web.1 connect=0ms service=1ms status=200 bytes=73930 protocol=http
```

This might be an expected warning since we have a cascade of forwards.

remlse:
> One thing from the logs:
> 
> ```
> 2023-04-10T17:50:22.913479+00:00 app[web.1]:    >> Parameter guard `category: Category` is forwarding: "No category called <zh-TW>".
> 
> 2023-04-10T17:50:23.046223+00:00 heroku[router]: at=info method=GET path="/static/styles/vendor_10880690442070639967.css" host=www-staging.rust-lang.org request_id=d7ad0c83-cb43-414f-81f6-33e4bafd572f fwd="24.7.113.20" dyno=web.1 connect=0ms service=1ms status=200 bytes=73930 protocol=http
> ```
> 
> This might be an expected warning since we have a cascade of forwards.

I think so, yes. Clicking through the website with the zh-TW locale, I can't find any route that's broken.

remlse:
I clicked around, making sure I'm hitting every route at least once. Everything seems fine, so the route ranking should be working at least.

Manishearth:
Hit the deploy button, will take a while for Heroku to actually pick it up and deploy it. I'll keep an eye out in the meantime.

Thanks for working on this!

PR #1811, opened 2023-04-21
Fix Cache-Control header

jdno:


Turbo87 (approved):

remlse:
Thanks! I borked this in the rocket v5 migration. I'm wondering if something like this would be more idiomatic:

```rust
use rocket::http::{hyper, Header};

header: Header::new(
    hyper::header::CACHE_CONTROL.as_str(),
    format!("max-age={max_age}"),
),
```

So we aren't hard coding the string at all.

jdno:
No worries about the migration, that can happen. 🙂 Sorry from my side for not writing a proper pull request description. 😅

I like the proposed code, much cleaner to rely on hyper than to hard-code the header name. Works well. 👍

PR #1861, opened 2023-09-02
fix mistake with the cargo add command

### LINKED ISSUE #1860: Mistake in getting started example project
### What needs to be fixed?
The example getting started example suggests the use of the command 
`cargo add ferris-says@0.2`
This however causes 

### Page Affected
https://www.rust-lang.org/learn/get-started

MilimTheTrueOne:
The cargo add command in the example is installs an older version of ferris-says, that causes the code not to compile if the cargo add command is used to install the dependency.

Closes  #1860.

### DIFF
--- a/locales/en-US/learn.ftl
+++ b/locales/en-US/learn.ftl
@@ -119,7 +119,7 @@ learn-dependencies-steps = <p>Let’s add a dependency to our application. You c
       <p>In this project, we’ll use a crate called <a href="https://crates.io/crates/ferris-says"><code>ferris-says</code></a>.
       <p>In our <code>Cargo.toml</code> file we’ll add this information (that we got from the crate page):</p>
       { $cargotoml }
-      <p>We can also do this by running <code>cargo add ferris-says@0.2</code>.</p>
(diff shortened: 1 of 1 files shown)


senekor:
Good catch. I would tend to replace it with `cargo add ferris-says@0.3` though. Otherwise stuff will break again without a mistake on our side if `ferris-says` releases a breaking version `0.4`, which `cargo add` would automatically choose.

Manishearth (approved):

Manishearth:
ah, this was missed during the upgrade, thanks!

Manishearth:
AIUI ferris-says isn't planning on doing further updates so I'm comfortable with this

MilimTheTrueOne:
A friend of mine had encountered this when trying out rust, so i just went and fixed it. Thanks for the quick response

willbchang:
The command is still not updated on the page: https://www.rust-lang.org/learn/get-started

image

senekor:
@willbchang the change isn't deployed yet. Deploys of the site are manual.

PR #1909, opened 2024-01-09
Use Rust 1.75's async fn in trait instead of #[async_trait]

dtolnay:
Announcement: https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -1,4 +1,5 @@
 use std::error::Error;
+use std::future::Future;
 use std::sync::Arc;
 use std::time::Instant;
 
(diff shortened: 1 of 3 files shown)


dtolnay:
Note there is no change to Cargo.toml because www-rust-lang-org does not depend directly on async-trait. This code was using `rocket::async_trait` via this, which puts it in scope in all modules:

https://github.com/rust-lang/www.rust-lang.org/blob/e89f3015b6481118ff4334d24d42cfd96d6cbc53/src/main.rs#L6-L7

Manishearth (approved):

Issue #1946, opened 2024-04-03
Bad font rendering in rust-social-wide SVG logos


onefifth:
### What needs to be fixed?
As reported [here](https://github.com/rust-lang/www.rust-lang.org/pull/1944#issuecomment-2033628302), rust-social-wide SVG logos do not display correctly when Fria Sans font is not present.

### Suggested Improvement
Convert the contained text to SVG paths.

PR #1947, opened 2024-04-03
Convert text in rust-social-wide to SVG path

onefifth:
Fixes bad text rendering when font is not present on client device.

Closes https://github.com/rust-lang/www.rust-lang.org/issues/1946


Urgau (approved):
Thanks you very much!

Looks perfect and works as expected (tested on my Rust PR).

Issue #483, opened 2018-11-30
Templates parts are accessible through some urls


meven:
# Content Bug Report

## Page(s) Affected
Whenever a category has a template it can be accessed through the route /<category>/<subject>
https://beta.rust-lang.org/tools/editors for instance

## What needs to be fixed?
Prevent the /<category>/<subject> route to access parts of the website.

PR #1981, opened 2024-06-02
Remove redundant 404 handlers

senekor:
I'm not sure why these existed in the first place. If they are removed, any requests to paths starting with "components/" correctly return (localized) 404 pages.

Maybe this was intended to prevent accidentally matching a template in the components directory as a standalone page? Even if such directory-based templating was a problem at some point, it doesn't seem to be anymore.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -151,16 +151,6 @@ fn baseurl(lang: &str) -> String {
     }
 }
 
-#[get("/components/<_file..>", rank = 1)]
-fn components(_file: PathBuf) -> Template {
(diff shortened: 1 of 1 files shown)


senekor:
The git history shows that the handler was added in https://github.com/rust-lang/www.rust-lang.org/pull/636 to solve https://github.com/rust-lang/www.rust-lang.org/issues/483.

After removing the hander, I cannot reproduce the original issue, so I think we can safely get rid of it.

Manishearth (approved):

PR #2010, opened 2024-07-09
Remove production/users page

senekor:
The data on this page was very outdated.

It also doesn't provide much value anymore, now that Rust is used very broadly in the industry.


senekor:
previously discussed [here](https://github.com/rust-lang/www.rust-lang.org/pull/2006#issuecomment-2212417375)

senekor:
Should I add a redirect from `production/users` to just `production`?

senekor:
I tried to be as thorough as possible with removing unused stuff. Even two dependencies were not needed anymore. The heavy lifting was done by grepping for "users". I also searched for "user-logos", because a few of them are referenced individually. I kept those and deleted all others.

senekor:
There is a visible change to the `/production` page:

<details>
<summary>before</summary>

</details>

<details>
<summary>after</summary>

</details>

Manishearth (approved):

PR #2083, opened 2024-10-29
Revert "Remove unrecognized platform report button"

alphachart:
Reverts rust-lang/www.rust-lang.org#2070


senekor:
You used this button previously to open [an issue here](https://github.com/rust-lang/www.rust-lang.org/issues/2038) but stopped replying to my questions. You can continue answering my questions on that issue so I can help fix your issue. I'm not going to reintruduce this button without a good reason, it caused a lot of spam issues.

PR #2106, opened 2025-01-14
Add redirect for wg-async adoption by lang

traviscross:
If WG-async is to be adopted by lang, let's add a redirect from where its page previously appeared on our website to where it will now.

See:

- https://github.com/rust-lang/team/pull/1639


traviscross:
The underlying change went in here, so this redirect now should also.

cc @ehuss

senekor (approved):

PR #2113, opened 2025-02-17
Deploy

ehuss:


ehuss:
Well, like a doofus I didn't bother to look at staging before posting this. For some reason, Heroku doesn't seem to be getting updated. Are there some logs we can see to see why it didn't deploy?

senekor:
Oh I did check staging and I thought it was there. Is this not it?

https://www-staging.rust-lang.org/governance/teams/launching-pad

senekor:
real website has it too:

https://www.rust-lang.org/governance/teams/launching-pad

ehuss:
Oh thanks! So the reason I'm confused is that https://www.rust-lang.org/governance still has the green "Working Groups" section, but https://github.com/rust-lang/www.rust-lang.org/pull/2105/files#diff-c9ffeb59f5244eeeaf523291e642be993d05fa0ef5e3103a2b88905f75e2a370 should have removed that. Did I miss something when updating the template?

image

senekor:
Ah, I see. I'll try to fix this.

senekor:
Oh maybe the new page is just shown because the team data was updated. I guess it is possible that the deployment failed completely. Nevermind.

ehuss:
Hm, strange! Let me know if there's anything I can do to help. I assume since it looks like the docker image is building that it is somewhere in the deployment phase.

ehuss:
@senekor Just checking if you've been able to see why the deployment didn't work?
If you don't have time right now, maybe I can bug someone on infra?

senekor:
Oh, I'm sorry. I should've said that I don't think I can help if it's something with the deployment. I don't really know anything about it and I don't have access to logs or whatever maybe useful to debug this.

ehuss:
Oh, no worries! I misunderstood. I'll poke some people.

senekor:
I thought maybe just doing a second deployment would help 😄 Turns out some of your commits aren't actually merged to the deploy branch yet?

https://github.com/rust-lang/www.rust-lang.org/pull/2116

ehuss:
Yea, #2114 was a followup PR, but wasn't related to the issue here. I was waiting to see if we could figure out what went wrong before trying another deploy. Dunno if trying again will work, but doesn't hurt to try.

senekor:
This is fixed now. The Heroku app had to be updated. It was running on Ubuntu 20.04, which was approaching its end of life.

ehuss:
Thanks!

PR #2130, opened 2025-04-02
Remove from governance page retired T-compiler WGs

apiraino:
As discussed with T-compiler leads, we are removing WGs that are not active anymore

The list of retired WGs matches that in the [forge documentation](https://forge.rust-lang.org/compiler/working-areas.html) (the second table)

cc @davidtwco @wesleywiser (for an ack)

thanks

### DIFF
--- a/src/teams.rs
+++ b/src/teams.rs
@@ -119,6 +119,16 @@ impl Data {
             .filter_map(|team| Some((team.name.clone(), team.subteam_of.clone()?)))
             .collect();
 
+        // Phased out T-compiler Working Groups
+        let phased_out_compiler_wgs = [
+            "wg-meta".to_string(),
+            "wg-nll".to_string(),
+            "wg-polymorphization".to_string(),
+            "wg-prioritization".to_string(),
+            "wg-self-profile".to_string(),
+            "wg-traits".to_string(),
+        ];
+
         self.teams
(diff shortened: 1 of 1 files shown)


senekor (changes requested):
Shouldn't this be changed in https://github.com/rust-lang/team/ ? We get the information from there and don't update it manually here. If these WGs updated accordingly in the team repo, they should automatically disappear from the website.

apiraino:
The question is legit. The rust-lang/team is the Source of Truth for Everything and I am not sure about the cascading consequences of deleting a subteam, specifically in terms of permissions changes to its members on repositories or other machinery of the infra. Are you confident that it is safe @senekor ?

senekor:
Well no, but I don't manage the teams repo. I just think you should open a PR there. Those people will know. In the unlikely case that the teams repo can't be adjusted in a way that makes the problem go away on the website automatically, I'm happy to help figuring out a general solution. But hard coding a list of defunct working groups in our source code doesn't sit right with me.

apiraino:
ok, will ask our T-infra friends and then be back :slightly_smiling_face:

senekor:
I will close this for now, on the assumption that the website shouldn't need any changes. Feel free to reopen in case of new information.

PR #2133, opened 2025-04-05
Deploy website

m-ou-se:


senekor (approved):

senekor:
It takes a moment to deploy, but from now on it should be automatic.

Issue #2146, opened 2025-04-30
"Get Started" button on the home page has double padding


Snapstromegon:
### Page(s) Affected

https://www.rust-lang.org/

### What needs to be fixed?

The "Get Started" button has a `pl3 pr3` class while it's already in a `ph3` container. This leads to it not being aligned with all other content on the page on narrow screens:

![Image](https://github.com/user-attachments/assets/6b61065e-7d1c-4dbd-a6bf-481b07a1ffb6)

![Image](https://github.com/user-attachments/assets/6c9e30e9-d41a-481a-936d-98ac8252395b)

### Suggested Improvement

Remove the `pl3 pr3` class from the wrapping div in https://github.com/rust-lang/www.rust-lang.org/blob/69781ef721035a9b913dc456ac90e055692518c4/templates/index.html.hbs#L10C48-L10C55

senekor:
I think it looks fine. The button is already big for the small text, no need to make it bigger. It's far away from the other buttons, alignment isn't an issue.

Snapstromegon:
The intention isn't to make the button bigger (that's just a side effect), but that it doesn't breaks the layout lines every other element seems to adhere to.

senekor:
Like I said I think it looks fine.

PR #2147, opened 2025-04-30
Remove doubled horizontal padding from "get started" button

Snapstromegon:
This removes the double horizontal padding from the "get started" button on the home page to align it with all other content and buttons on the page

fixes: #2146


senekor:
I think it looks fine. The button is already big for the small text, no need to make it bigger. It's far away from the other buttons, alignment isn't an issue.

PR #2152, opened 2025-05-17
Include marker and unknown teams with a website section

Kobzol:
Before, they were not shown on the website, even though they had a website section in `team`. The current main motivation for this is to include GSoC mentors and contributors on the website, under the launching pad.

I think that the only other team other than GSoC mentors that is a marker team and has a website section now is https://github.com/rust-lang/team/blob/c4c9de724db63ea7479da976f3ebbcb9b34970ed/teams/alumni.toml. But since they don't have a parent team, and are not a top level team, I suppose that they won't be shown on the website?

### DIFF
--- a/src/teams.rs
+++ b/src/teams.rs
@@ -31,6 +31,8 @@ pub struct PageData {
     subteams: Vec<Team>,
     wgs: Vec<Team>,
     project_groups: Vec<Team>,
+    // Marker teams and other kinds of groups that have a website entry
+    other_teams: Vec<Team>,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


senekor (approved):

senekor:
The gap between the mentors team and the ones above is bigger, because they are in different sections. I guess it's not a big deal?

image

image

Kobzol:
Well, it's kind of intended, because they are indeed different sections, it's just that it's not really shown on the website and they have the same color. There should be the same gap between WGs and project groups, so I think that it's fine.

Issue #2165, opened 2025-06-13
Confusing "log" icon


apiraino:
### Page(s) Affected
https://www.rust-lang.org/what/cli

<details><summary>Screenshot</summary>
<p>

![Image](https://github.com/user-attachments/assets/e0854a0d-aaa5-4906-916a-313a099191c9)

</p>
</details> 

### What needs to be fixed?
I am not English mother tongue so it's not clear to me the connection between *logging* (the act of recording output from software to inspect what's happening) and a *log* (a piece of wood).

### Suggested Improvement
Representing concepts in a way that are familiar to a restricted culture group is a well-known UI issue. Perhaps a more relevant icon could be found to represent *data logging*?

Thanks

senekor:
What about a seismograph? could be a nice symbol of continuous collection of data.

apiraino:
Based on a bit of sampling of companies (or projects) offering data logging[^1] it seems that a common visual representation are histograms and graphs (so basically the seismograph idea could fit) or some text resembling the [Common Log Format](https://duckduckgo.com/?q=common%20log%20format&ia=images&iax=images) optionally with a magnifying lens over them.

not a qualified visual designer, just throwing ideas 🙂 

[^1]: https://www.datadoghq.com, https://matomo.org, https://plausible.io, https://duckduckgo.com/?q=server+logs+analysis+clipart&iar=images

PR #2187, opened 2025-09-02
Change CLI logging SVG

senekor:
The wooden logs were not an intuitive symbolism for non-native English speakers.

(This doesn't affect the PR for static rendering #2174)

closes: #2165

image


senekor:
CI is failing because of a new clippy warning, I think we should ignore it so we don't create a conflict with the static rendering PR.

PR #2195, opened 2025-09-19
deploy the static website

sysrex:
Just changes the base url in preparation for the DNS change.


.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
senekor, line 1: Please drop this (and the added/deleted files) from the history. `.idea` belongs in the user's global gitignore.

senekor (changes requested):

Manishearth (approved):

PR #2202, opened 2025-09-26
Add archived teams to the website

Kobzol:
This PR adds a new section to the Governance page which links to a single page that contains all archived teams on one place. Especially the design of the archived page section/link on the main governance page is kind of wobbly, I'm open to bikeshedding :)

Currently, we have 40 archives teams, and 9 of them did not have `website_data` entry. I modified the website to show at least something for these teams, so that we don't miss out on any archived teams. But an alternative would be to add this data to the `team` repository. Let me know what you think.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -740,7 +740,7 @@ dependencies = [
 [[package]]
 name = "rust_team_data"
 version = "1.0.0"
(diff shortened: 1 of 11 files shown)


Kobzol:
Done. I also had to rebase.

PR #2208, opened 2025-10-07
Make logos available under `/logos`

Kobzol:
I noticed on my GitHub page that the Rust logo stopped working. I was using a link to https://www.rust-lang.org/logos/rust-logo-16x16-blk.png, but that is now uder https://www.rust-lang.org/static/logos/rust-logo-16x16-blk.png.

This PR makes the logos available also under the original URL. Note that we can't use HTML redirects here, as it's a static file asset (an image).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -100,6 +100,11 @@ fn main() -> anyhow::Result<()> {
         base_url,
     };
     ctx.copy_asset_dir("static", "static")?;
+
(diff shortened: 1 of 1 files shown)


senekor:
Why exactly do we need this? What is "your GitHub page" and can you update the link there?

senekor (approved):

PR #2211, opened 2025-10-11
Add a page with everyone in the Rust Project, including alumni

Kobzol:
The recent addition of archived teams finally included more alumni on the website. But I think that we're still missing some high-level view. So I thought - why not show *everyone*? The whole Rust Project (although I didn't call it like that on the website, as we don't really have a precise definition of that yet) on a single page, including all alumni.

https://github.com/user-attachments/assets/8a04f34b-7d89-42e3-873c-167c5bae1c3e

### DIFF
--- a/locales/en-US/governance.ftl
+++ b/locales/en-US/governance.ftl
@@ -11,6 +11,8 @@ governance-rfc-blurb = Each major decision in Rust starts as a Request for Comme
 governance-teams-header = Teams
 governance-wgs-header = Working Groups
 
+governance-rust-project-link = Show all Rust team members
+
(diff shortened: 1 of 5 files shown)


apiraino:
@Kobzol Listing **everyone** having an entry in our `team` repository under the `./people` directory is a bit different than being active member of a Rust team, right?

Kobzol:
This PR added a page with all members of all Rust teams in the `team` repo (not just actual teams, but also WGs, etc.). We don't yet have a clear definition of what the Rust Project is, but this is one of the definitions that we have used so far (AFAIK).

Note that the page splits active members and alumni.

PR #2213, opened 2025-10-24
Add a separate page for each Rust team member

Kobzol:
Create a separate page for each Rust Team member, which lists the teams they are a member (or alumni) of. This can be helpful to quickly show that someone is a part of the Project, e.g. when looking for a job.

The commit that adds Rayon reduces the render time from ~10s to ~2s on my PC, it can be helpful when iterating on the website. But if you don't like including Rayon, I can get rid of it.

image

### DIFF
--- a/templates/governance/group.html.hbs
+++ b/templates/governance/group.html.hbs
@@ -2,34 +2,34 @@
 
 {{#with data.team as |team|}}
     <section class="white">
-        {{> governance/group-team team=team zulip_domain=data.zulip_domain baseurl=baseurl}}
+        {{> governance/group-team team=team zulip_domain=data.zulip_domain baseurl=../baseurl}}
     </section>
 {{/with}}
 {{#if data.subteams}}
     <section class="purple">
         {{#each data.subteams as |team|}}
-            {{> governance/group-team team=team zulip_domain=data.zulip_domain baseurl=baseurl}}
+            {{> governance/group-team team=team zulip_domain=data.zulip_domain baseurl=../baseurl}}
         {{/each}}
     </section>
 {{/if}}
 {{#if data.project_groups}}
     <section class="green">
         {{#each data.project_groups as |team|}}
-            {{> governance/group-team team=team zulip_domain=data.zulip_domain baseurl=baseurl}}
+            {{> governance/group-team team=team zulip_domain=data.zulip_domain baseurl=../baseurl}}
         {{/each}}
     </section>
 {{/if}}
 {{#if data.wgs}}
     <section class="green">
         {{#each data.wgs as |team|}}
-            {{> governance/group-team team=team zulip_domain=data.zulip_domain baseurl=baseurl}}
+            {{> governance/group-team team=team zulip_domain=data.zulip_domain baseurl=../baseurl}}
         {{/each}}
     </section>
 {{/if}}
 {{#if data.other_teams}}
     <section class="green">
         {{#each data.other_teams as |team|}}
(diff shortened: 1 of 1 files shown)


senekor:
> The value is just that I wanted to have a shared logic for it, rather than risking that it will diverge across different pages.

I don't agree with the reasoning here. There's technically a lot of duplication in the templates. Creating a separate rust helper for each occurrence of duplication would lead to a lot of unnecessary complexity.

What about [this](https://github.com/rust-lang/www.rust-lang.org/blob/5443ac1626ebe66a42686f89e5963e73371f0714/templates/governance/archived-team.html.hbs#L4)? Looks like you forgot to update a place where the "team anchor helper" should've been used? If so, the helper actually made the risk of divergence worse. Somebody might update the helper, thinking that should take care of everything, and forget to search for places where the helper wasn't used.

I think it's better to just copy these snippets everywhere. When somebody updates a hardcoded URL path, it's very obvious that searching for other occurrences is necessary. And when looking at the template, you can see what it does instead of having to understand what a handlebars helper is and dig through the Rust code.

The two existing helpers contain logic that would actually be tricky and brittle to implement in the templating language. These proposed helpers don't have logic, they're just snippets. I'd rather do without them.

Actually, there is a tiny bit of logic in the `PersonLinkHelper` in the form of making the github username lowercase. That only seems to be necessary because the same thing is done when rendering the pages. Can we just get rid of both and render the github usernames with original casing in the path? Or what's the reason to force it to lowercase?

Kobzol:
> What about [this](https://github.com/rust-lang/www.rust-lang.org/blob/5443ac1626ebe66a42686f89e5963e73371f0714/templates/governance/archived-team.html.hbs#L4)? Looks like you forgot to update a place where the "team anchor helper" should've been used? If so, the helper actually made the risk of divergence worse. Somebody might update the helper, thinking that should take care of everything, and forget to search for places where the helper wasn't used.

Yes, it should have also been used there, which is IMO an argument for having the helper, so that we don't forget to update multiple places if the heading changes. But I don't feel too strongly about this - for me as a developer of the PR, it would ofc be the easiest to just copy-paste everything, I wanted to make the code easier for you to maintain into the future, which is why I introduced the helpers :laughing: But if you see this differently, then there's no point in it, I'll remove the helpers.

> Actually, there is a tiny bit of logic in the PersonLinkHelper in the form of making the github username lowercase. That only seems to be necessary because the same thing is done when rendering the pages. Can we just get rid of both and render the github usernames with original casing in the path? Or what's the reason to force it to lowercase?

The URLs are case sensitive. Some GitHub usernames are quite complex regarding casing, so I thought that normalizing them might be a better idea, if someone wanted to generate the links manually (e.g. my username on github is `Kobzol`, but I usually just write `kobzol`). If we don't think that this use-case is required, we can just depend on people finding out about these links through the website itself, in which case we don't need to normalize.

Kobzol:
Removed the helpers and lowercasing, and added a link to the person page when clicking on avatars (except for the person page itself, there the avatar links to GitHub).

senekor:
I just noticed that the language is not preserved when clicking on the new person links. e.g.  when I'm here:

http://localhost:8000/fr/governance/teams/leadership-council/

and click on any person, the `fr/` is removed and I'm on the English version of the website. The language should be preserved.

Kobzol:
Sigh, Handlebars doesn't report that the `baseurl` variable was unset :( I don't like this templating system.

Kobzol:
Let me know if there's anything else I can change! :)

senekor:
Oops, sorry. I didn't notice the new commit.

PR #2227, opened 2025-11-04
Sort team members on the website

Kobzol:
Before, we were using the original order from `team`.

This was reported in a (private) compiler thread on Zulip.

### DIFF
--- a/src/teams.rs
+++ b/src/teams.rs
@@ -189,7 +189,7 @@ impl RustTeams {
                     b.is_lead.cmp(&a.is_lead)
                 } else if a.roles.len() != b.roles.len() {
                     // Then by role count (more roles before less roles)
-                    b.roles.cmp(&a.roles)
+                    b.roles.len().cmp(&a.roles.len())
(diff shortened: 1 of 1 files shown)


src/teams.rs
```diff
@@ -181,6 +181,41 @@ impl RustTeams {
...
 
+        fn sort_members(members: &mut [TeamMember]) {
+            members.sort_by_key(|member| member.github.to_lowercase());
+            members.sort_by(|a, b| {
+                // First, sort by lead status (lead before non-lead)
+                if a.is_lead != b.is_lead {
+                    b.is_lead.cmp(&a.is_lead)
+                } else if a.roles.len() != b.roles.len() {
+                    // Then by role count (more roles before less roles)
+                    b.roles.cmp(&a.roles)
```
senekor:
```suggestion
                    b.roles.len().cmp(&a.roles.len())
```

Kobzol:
Committed the suggestion.

PR #2232, opened 2025-11-08
Add the total numbers of members and alumni

yotamofek:
My partner asked me roughly how many people are "official" rust team members, and I couldn't say, so thought it might be a good addition to the "all team members" page.

Not sure about the placement and whether others will think it's a good idea too.

r? @Kobzol as the person who added this page in the first place :)


Kobzol:
I was also thinking that displaying the count would be nice. 👍

PR #2237, opened 2025-11-14
Add a funding page that lists Project members with GitHub Sponsors

Kobzol:
This PR adds a dedicated Funding page that lists all active Project members that have GitHub Sponsors enabled and have opted into it being publicized through [`team`](https://github.com/rust-lang/team/pull/2071). The order of people is randomized using JavaScript on every page load, to reduce ordering bias.

Maybe we could even add a dedicated "Funding" link in the top-level navbar?

image

image

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -106,6 +106,7 @@ pub struct CSSFiles {
 #[derive(Serialize, Debug)]
 pub struct JSFiles {
     tools_install: String,
+    funding_shuffle: String,
 }
 
 #[derive(Serialize, Debug)]
@@ -125,6 +126,7 @@ pub fn compile_assets(
     let fonts_css_file = compile_sass(root_dir, out_dir, "fonts", base_url)?;
     let vendor_css_file = concat_vendor_css(root_dir, out_dir, vec!["tachyons"])?;
     let tools_install_js = build_js_file(root_dir, out_dir, "tools-install")?;
+    let funding_shuffle_js = build_js_file(root_dir, out_dir, "funding-shuffle")?;
 
     Ok(AssetFiles {
         css: CSSFiles {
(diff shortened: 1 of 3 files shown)


Kobzol:
> This looks pretty good. I don't like the handling of the custom script though. This is basically adding a hook where every page can inject it's custom javascript, right? Why not just write it inline inside the "page" block? I checked if there's a precedent for this. `tools-install.js` is another custom script that only applies to one page. But it seems to me that's handled in a complicated way too... it get's another generated name "vendor_<random_num>.js". I can't see a reason why this is done. I would prefer if we put all scripts in `static/scripts/` and reference them explicitly with the same name as the source file in HTML templates. What do you think?

The main reason why I added the script section is because when I added the script to the end of the `page` inline, it broke the logic I added in https://github.com/rust-lang/www.rust-lang.org/pull/2231, because the last `section` was no longer the last child of its parent, as there now was the `<script>` :laughing: In general, I think it's slightly better if JS scripts are at the very end of the page, rather than being in some div inside the page.

I'm fine with adding the logic into an external file, although it makes the logic a bit more distributed. When it's in the same file, it's easier to see that you should update the script when the HTML IDs/classes change, for example.

senekor:
Could it be that the generated name for the `tools-install.js` file is intended to break caching? To force a reload of the file in case it changed?

Kobzol:
> Could it be that the generated name for the `tools-install.js` file is intended to break caching? To force a reload of the file in case it changed?

Yup, pretty sure that's the reason.

Created a separate JS script for the funding page.

I also removed the "app" JS file, since I think that it doesn't make sense to combine the funding JS (as the scripts might not be self-contained and shouldn't run on arbitrary pages) and the tools installation JS, and instead created a separate JS file for both.

PR #2245, opened 2025-11-25
Add Funding page navbar link

Kobzol:
As a part of the effort to promote it (https://rust-lang.zulipchat.com/#narrow/channel/392734-council/topic/Publicizing.20our.20funding.20page/with/560030032, https://github.com/rust-lang/blog.rust-lang.org/pull/1751).

image

We should probably translate it into all current languages, since an untranslated navbar item will probably look quite weird?

Marking as a draft, to sync it with the blog post and other changes.

### DIFF
--- a/locales/de/common.ftl
+++ b/locales/de/common.ftl
@@ -20,6 +20,7 @@ nav-install = Installieren
 nav-learn = Lernen
 nav-tools = Werkzeuge
 nav-governance = Leitung
+nav-funding = Finanzierung
 nav-community = Community
 nav-blog = Blog
 choose-language = Sprache
--- a/locales/en-US/common.ftl
+++ b/locales/en-US/common.ftl
@@ -31,6 +31,7 @@ nav-install = Install
 nav-learn = Learn
 nav-tools = Tools
 nav-governance = Governance
+nav-funding = Funding
 nav-community = Community
 nav-blog = Blog
 nav-playground = Playground
--- a/locales/es/common.ftl
+++ b/locales/es/common.ftl
@@ -28,6 +28,7 @@ nav-install = Instala
 nav-learn = Aprende
 nav-tools = Herramientas
 nav-governance = Gobernanza
+nav-funding = Financiación
 nav-community = Comunidad
 nav-blog = Blog
 nav-playground = Playground
--- a/locales/fr/common.ftl
+++ b/locales/fr/common.ftl
@@ -28,6 +28,7 @@ nav-install = Installer
 nav-learn = Apprendre
 nav-tools = Outils
 nav-governance = Gouvernance
+nav-funding = Financement
 nav-community = Communauté
 nav-blog = Blog
 nav-playground = Bac à sable
--- a/locales/it/common.ftl
+++ b/locales/it/common.ftl
@@ -28,6 +28,7 @@ nav-install = Installa
 nav-learn = Impara
 nav-tools = Strumenti
 nav-governance = Governance
+nav-funding = Finanziamento
 nav-community = Community
 nav-blog = Blog
 nav-playground = Playground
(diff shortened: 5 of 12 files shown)


lqd:
are we good here?

Kobzol:
When I originally proposed this, Remo [noted](https://rust-lang.zulipchat.com/#narrow/channel/392734-council/topic/Publicizing.20our.20funding.20page/near/560086130) that the Funding page was a bit bare bones back then for it to be linked in the navbar. I kind of agree.

Right now, the funding page is essentially the same, however, the RFMF is finally picking up steam, and we will now have a very important link on the funding page that will lead to a donate button to give money to Rust. I think that is important enough for us to include the funding link in the navbar.

@senekor What do you think?

lqd:
Same question to @Manishearth

Manishearth (approved):
I'm fine with this.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/www.rust-lang.org/2245/84e18218f1265c1f03f56f91905a98b84098c506..88d3e3b6e54b55ade959f4bdb4540720212f4733)*

Kobzol:
Since it is in the navbar, I think that we should translate this into other languages. I know that there is a freeze on new translations, but having untranslated text in the navbar feels.. meh. Do you agree?

senekor:
I'm also fine with it once the RFMF is live. We should make the RFMF the more prominent option though. Now it's just a link as part of a sentence. A big, colorful DONATE button or whatever seems best. Explanations of other options can be moved below that.

I also agree with the translation issue. I wish we had a sustainable solution to keep the translations up to date. Hot take: what if we do a call for action among rust project members to step up and help maintain translations for their native languages? Volunteers will be explicit (new translation teams per language?) with an expectation to help out long-term. If there are no volunteers for that language or such a team becomes inactive, we simply remove the language from the list of selectable options.

At the end of the day, helping with translations isn't that much work. We just need somebody who's willing to be pinged every couple months and write a couple sentences.

Manishearth:
> Since it is in the navbar, I think that we should translate this into other languages. I know that there is a freeze on new translations, but having untranslated text in the navbar feels.. meh. Do you agree?

Yes. We can ask around for translations + potentially use some machine translations. It's easier for single words since you can look it up in a dictionary and be sure it's the right sense of the word.

lqd:
Eventually, we'd need to also translate the funding page itself, not just the link. And even then GH itself isn't translated so every piece of traffic traversing the funding page from and to github will have to involve English, i.e. everything about github sponsors that we want to enable.

I'll try to translate the single link in the languages supported later today, and ask for confirmation to the project members from these countries.

Kobzol:
Rémy gathered a bunch of translations for the link, which is great. As he correctly pointed out, the funding page itself, and GitHub Sponsors, is untranslated anyway, so it's all just best-effort.

I think that this is now ready to be merged, we can always add more translations in the future, and we will keep updating the funding page.

Creating a team for translations seems like a worthy effort, but I'm not sure if we have the bandwidth at the moment, and I definitely do not want for funding work to be blocked on that =D

senekor:
Wouldn't "donate" maybe be a better term than "funding" ? I know that in English, it can kind of mean both depending on the context. But in German, the term "Finanzierung" makes less sense IMO. I would expect a "Finanzierung" page to maybe have transparancy information about who is giving money to Rust. Not information about how to donate yourself. In German, "Spenden" i.e. "donate" would make more sense to me.

(German isn't actually a selectable language on the website, but the file is still in the repo, which made me think about this.)

Kobzol:
We talked about this with Rémy :laughing: ... yes, in retrospect I think that donate is in fact a better name for the link :man_facepalming:

Kobzol:
Since specifically German is not used on the website right now (not sure why, actually), can we merge what we have now, to put something in the navbar that links to the Funding page, as it is now essentially unlinked? We can always improve the wording later.

senekor (approved):

PR #2260, opened 2026-01-12
Fix alignment on learn page

### LINKED ISSUE #2256: Alignment problem on the non-english versions of "Learn" page with link to a non-official translation of The Book
### Page(s) Affected

https://rust-lang.org/fr/learn/ and other non-english versions of the learn page.

### What needs to be fixed?

The alignment between the button "LIRE LE BOOK" and the link "Lire la traduction non-officielle !" is off.
At the time the second text was added, the button was centered (see screenshots at PR #1370).

### Suggested Improvement

Centering the button is a first option.
A second option is to align the link to the left.
Which option seems right ?

senekor:
closes #2256

### DIFF
--- a/templates/funding.html.hbs
+++ b/templates/funding.html.hbs
@@ -7,7 +7,7 @@
     </a>
     <div class="flex flex-column">
       <a href="{{baseurl}}/governance/people/{{member.github}}">{{member.name}}</a>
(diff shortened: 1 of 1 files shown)


apiraino:
thanks for fixing the regression I introduced :)

PR #2261, opened 2026-01-16
feat: Add an `All Teams` page

### LINKED ISSUE #2259: Add a `View All Teams` page
### Summary
It would be amazing if there was a page, similar to the `view all team members` page that would show a list of all the teams working under the rust project. It could either be a simple list, or a full one with names of all the team members (like they are currently listed).

### Motivation
It is really hard to find a specific team when you're looking for it. So often, I'll end up having to look through all the sections (Launching Pad, Compiler, etc) to find the specific team that I'm looking for. This would make it much easier to do so.

### Drawbacks
I really can't think of many. One could be that it would increase the footprint of the site. 

I do not know about other decisions on this issue (if there even have been any). 

### Maintenance
This design should not be a maintainance burden IMO. We could use the existing lists, and simply flatten them out. I'm not sure about what the doc comment here meant by 'who will maintain it'. Please feel free to enlighten me :)

### Unresolved Questions
N/A (for now)

JayanAXHF:
Closes #2259

### DIFF
--- a/templates/governance/all-teams.html.hbs
+++ b/templates/governance/all-teams.html.hbs
@@ -24,14 +24,8 @@
   <div class="w-100 mw-none mw-8-m mw9-l center f2">
     <h2>{{fluent "governance-all-teams-archived-title"}}</h2>
     <p>
-      {{#fluent "governance-all-teams-archived-intro"
-
-
-      }}{{#fluentparam
-      "count"
-
-
-      }}{{len data.archived}}{{/fluentparam}}{{/fluent}}
+      {{#fluent "governance-all-teams-archived-intro" }}{{#fluentparam "count"}}
+      {{len data.archived}}{{/fluentparam}}{{/fluent}}
     </p>
   </div>
   <div>


Kobzol:
Looks nice! I didn't find any link to the page though?

I think that we could replace the Archived teams section on the governance page with a link to this page, which will simply list all teams (both active and archived), same as the all people page lists both active contributors and alumni. If we then make the "Archived teams" header on your page clickable (really it can just have an ID), it will be possible to link to the archived teams. We can thus join the two pages together, and figure out some better way to link to this combined page from the governance page.

src/redirect.rs
```diff
@@ -37,9 +37,13 @@ pub static PAGE_REDIRECTS: &[(&str, &str)] = &[
...
     ),
     // miscellaneous
-    ("governance/teams", "governance#teams"),
     ("governance/wgs", "governance#working-groups"),
     ("governance/all-team-members.html", "governance/people"),
+    (
+        "governance/archived-teams.html",
+        "governance/teams#archived-teams",
+    ),
+    ("governance/all-teams.html", "governance/teams"),
```
senekor: Can you explain the changes to the redirects here?

src/render.rs
```diff
@@ -217,25 +219,36 @@ pub fn render_governance(
...
+    // for_all_langs("governance/archived-teams.html", |dst_path, lang| {
+    //     render_ctx
+    //         .page(
+    //             "governance/archived-teams",
+    //             "governance-archived-teams-title",
+    //             &archived_teams_data,
+    //             lang,
+    //         )
+    //         .render(dst_path)
+    // })?;
```
senekor: remove unused code

src/teams.rs
```diff
@@ -252,7 +265,6 @@ impl RustTeamData {
                 active.extend(team.members.iter().map(|m| (m.github_id, m)));
             }
         }
-
```
senekor: revert this whitespace change

src/teams.rs
```diff
@@ -433,17 +445,23 @@ pub struct PageData {
...
 }
 
-#[derive(Serialize)]
-pub struct ArchivedTeams {
-    teams: Vec<Team>,
-}
+// #[derive(Serialize)]
+// pub struct ArchivedTeams {
+//     teams: Vec<Team>,
+// }
```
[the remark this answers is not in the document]
JayanAXHF: Fixed all three in [897a83d](https://github.com/rust-lang/www.rust-lang.org/pull/2261/commits/897a83d809fbed68cb52b2650c4afb83e39ca9d6)

src/redirect.rs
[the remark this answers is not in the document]
JayanAXHF: Oh that's my bad, i'll remove it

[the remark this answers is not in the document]
JayanAXHF: Ok so i removed the redirect from all-teams.html, and the other one redirects from the old archived teams page to the new one

src/teams.rs
JayanAXHF: Done!

JayanAXHF:
> I suppose the template `archived-teams.html` is not used anymore. In that case, it should be deleted as well.

Done!

templates/governance/all-teams.html.hbs
```diff
@@ -0,0 +1,174 @@
+{{#*inline "active"}}
```
senekor: This looks like it's copied from `group-team.html.hbs`. Can you extract the shared part an reuse it so we don't have that much duplication?

JayanAXHF:
Made the Changes!

locales/en-US/governance.ftl
```diff
@@ -39,6 +40,13 @@ governance-all-team-members-title = All Rust team members
 governance-all-team-members-intro = This section lists the { $count } members of currently active Rust teams.
 governance-all-team-members-alumni-intro = This section lists our { $count } team alumni.
 
+## governance/all-teams.hbs
+governance-all-teams-title = All Rust Teams
+governance-active-teams-title = Active Rust Teams
+governance-all-teams-intro = This section lists the { $count } teams currently active in the Rust project.
+governance-all-teams-archived-title = Archived Rust Teams
+governance-all-teams-archived-intro = This section lists { $count } archived teams that are no longer active. We want to thank all past members for their invaluable contributions!
```
senekor:
```suggestion
governance-all-teams-archived-intro = This section lists the { $count } archived teams that are no longer active. We want to thank all past members for their invaluable contributions!
```

Add "the" for consistency with the section title of active teams?

senekor (changes requested):

templates/governance/all-teams.html.hbs
```diff
@@ -0,0 +1,44 @@
...
+    <h2 class="f3 f1-ns mt0">
+      {{fluent "governance-active-teams-title"}}
+    </h2>
+    <p>
+      {{#fluent "governance-all-teams-intro" }}{{#fluentparam "count" }}{{len data.active}}{{/fluentparam}}{{/fluent}}
+    </p>
+  </div>
+  <div>
+    {{#each data.active as |team|}}
+    {{> governance/group-team team=team baseurl=../baseurl zulip_domain=../data.zulip_domain }}
```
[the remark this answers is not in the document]
JayanAXHF: Right, fixed!

senekor (approved):
good stuff, thanks!

PR #2277, opened 2026-04-06
fix: update learn page embedded book link

### LINKED ISSUE #2274: 'Embedded book' link seems to be outdated at /learn page
### What needs to be fixed?
Knowing little about available embedded book options, I found myself at 
https://doc.rust-lang.org/stable/embedded-book/  by following yellow 'Embedded Book' button/link from https://rust-lang.org/learn/.
It turned out to be an oldest STM F3 book out of 3 available, which is unfortunate.

### Page(s) Affected
https://rust-lang.org/learn/
Yellow 'Embedded Book' link.

### Suggested Improvement
I suggest changing that link to https://docs.rust-embedded.org/discovery/ or any other link, pointing to all the available material.
Cheers.

abhijeetnardele24-hash:
## Summary
- update the Learn page Embedded Book button to point to the current embedded Rust book
- replace the outdated doc.rust-lang.org embedded-book link

Closes #2274

### DIFF
--- a/templates/learn/index.html.hbs
+++ b/templates/learn/index.html.hbs
@@ -120,7 +120,7 @@
           </div>
 
           <div class="pt4 pt3-l flex flex-column flex-row-l items-center-l mb4-l">
-            <a href="https://doc.rust-lang.org/embedded-book"
+            <a href="https://docs.rust-embedded.org/book/"
                class="button button-secondary mw6-l w-100">
               {{fluent "learn-domain-embedded-button"}}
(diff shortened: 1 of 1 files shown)


senekor:
These two links point to the same book. Please don't post LLM-generated garbage PRs, this only wastes reviewer's time.

(One of the two URLs technically lags behind a little, as that version is only updated whenever a new stable version of Rust is released. But that is fine and not what the linked issue is about.)

PR #2278, opened 2026-04-06
fix: add working group redirects

### LINKED ISSUE #2268: Add redirects for working group pages
Some old working-group URLs on rust-lang.org return 404s instead of redirecting to the correct location.

For example:

https://rust-lang.org/governance/wgs/wg-secure-code -> 404

but it should redirect to:

https://rust-lang.org/governance/teams/#team-wg-secure-code

abhijeetnardele24-hash:
## Summary
- add redirects for the current launching-pad working group pages that used to live under /governance/wgs
- point the old WG URLs at the corresponding anchors on the all teams page

Closes #2268


senekor:
Please explain how you determined that these redirects are the right ones to fix the issue.

PR #2281, opened 2026-04-09
feat: add copy button on install link to improve Accessibility

### LINKED ISSUE #2280: Add copy button in Install link to improve Accessibility
### Summary

https://github.com/user-attachments/assets/f564ece3-fc85-45ee-9632-fda9b0da410b

When only using the keyboard to navigate throughout the Install page, it is very hard to copy the rust  installation command as there is no copy button present, This also makes the page less [accessible](https://www.w3.org/WAI/fundamentals/accessibility-intro/).

### Motivation

I'm working in the Accessibility team at [djangonauts](https://djangonaut.space/comms/2026/02/19/session-6-team-introductions/) and my work is to make django more accessible to the people with special abilities. And I'm trying to learn Rust and when I go to the installation page, I realized that the rust Installation command in the install page is not accessibility friendly and someone using accessibility tools can't able to copy the command to install rust. So to make it accessibility friendly we can add a copy button, so that people who are using accessibility tools can copy the installation command.

### Drawbacks

I don't think there is any drawback to add this copy button.

In my view, there are no other alternatives at the moment, but I would be happy to hear any new suggestions or ideas.
### Maintenance

I think it is very easy to maintain.

Mr-Sunglasses:
fix: #2280 

Add a copy button in installation command to make it more accessibility friendly.

https://github.com/user-attachments/assets/a9a95764-a232-4d67-a987-c2fe5cf29517

### DIFF
--- a/static/scripts/tools-install.js
+++ b/static/scripts/tools-install.js
@@ -1,184 +1,287 @@
 var platform_override = null;
 
 function vis(elem, apply) {
-    ["db", "di", "dn"].forEach(c => {
-        if (c === apply) {
-            elem.classList.add(c);
-        } else {
-            elem.classList.remove(c);
-        }
-    });
+  ["db", "di", "dn"].forEach((c) => {
+    if (c === apply) {
+      elem.classList.add(c);
+    } else {
+      elem.classList.remove(c);
+    }
+  });
 }
 
 function detect_platform() {
-    "use strict";
+  "use strict";
 
-    if (platform_override) {
-        return platform_override;
+  if (platform_override) {
+    return platform_override;
+  }
+
+  var os = "unknown";
+
+  if (navigator.platform == "Linux x86_64") {
+    os = "unix";
+  }
+  if (navigator.platform == "Linux i686") {
+    os = "unix";
+  }
+  if (navigator.platform == "Linux i686 on x86_64") {
+    os = "unix";
+  }
+  if (navigator.platform == "Linux aarch64") {
+    os = "unix";
+  }
+  if (navigator.platform == "Linux armv6l") {
+    os = "unix";
+  }
+  if (navigator.platform == "Linux armv7l") {
+    os = "unix";
+  }
+  if (navigator.platform == "Linux armv8l") {
+    os = "unix";
+  }
+  if (navigator.platform == "Linux ppc64") {
+    os = "unix";
+  }
+  if (navigator.platform == "Linux mips") {
+    os = "unix";
+  }
+  if (navigator.platform == "Linux mips64") {
+    os = "unix";
+  }
+  if (navigator.platform == "Linux riscv64") {
+    os = "unix";
+  }
+  if (navigator.platform == "Mac") {
+    os = "unix";
+  }
+  if (navigator.platform == "Win32") {
+    os = "win";
+  }
+  if (navigator.platform == "FreeBSD x86_64") {
+    os = "unix";
+  }
+  if (navigator.platform == "FreeBSD amd64") {
+    os = "unix";
+  }
+  if (navigator.platform == "NetBSD x86_64") {
+    os = "unix";
+  }
+  if (navigator.platform == "NetBSD amd64") {
+    os = "unix";
+  }
+
+  // I wish I knew by now, but I don't. Try harder.
+  if (os == "unknown") {
+    if (navigator.appVersion.indexOf("Win") != -1) {
+      os = "win";
     }
-
-    var os = "unknown";
-
-    if (navigator.platform == "Linux x86_64") {os = "unix";}
-    if (navigator.platform == "Linux i686") {os = "unix";}
-    if (navigator.platform == "Linux i686 on x86_64") {os = "unix";}
-    if (navigator.platform == "Linux aarch64") {os = "unix";}
-    if (navigator.platform == "Linux armv6l") {os = "unix";}
-    if (navigator.platform == "Linux armv7l") {os = "unix";}
-    if (navigator.platform == "Linux armv8l") {os = "unix";}
-    if (navigator.platform == "Linux ppc64") {os = "unix";}
-    if (navigator.platform == "Linux mips") {os = "unix";}
-    if (navigator.platform == "Linux mips64") {os = "unix";}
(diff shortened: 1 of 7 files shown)


senekor:
Please revert the whitespace changes so this can be reviewed more easily.

Mr-Sunglasses:
> Please revert the whitespace changes so this can be reviewed more easily.

Hey @senekor 👋🏻 , I've reverted the whitespace changes, thanks for the suggestion.

templates/components/tools/rustup.html.hbs
```diff
@@ -1,7 +1,19 @@
...
+    <div class="pre-copy-wrap">
+      <pre><code class="db w-100 copyable">curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh</code></pre>
+      <button class="copy-code-btn" type="button" aria-label="Copy install command to clipboard">
+        <svg class="copy-icon" xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true" focusable="false">
+          <rect x="9" y="2" width="6" height="4" rx="1"/>
+          <path d="M8 4H6a2 2 0 0 0-2 2v14a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V6a2 2 0 0 0-2-2h-2"/>
+        </svg>
+        <svg class="check-icon" xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2.5" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true" focusable="false">
+          <polyline points="20 6 9 17 4 12"/>
+        </svg>
```
senekor: My instinct would be to move these into separate `.svg` files.

static/scripts/tools-install.js
```diff
@@ -177,8 +177,57 @@ function check_initial_override() {
...
+                    btn.classList.remove('copied');
+                }, 2000);
+            }
+
+            if (navigator.clipboard && navigator.clipboard.writeText) {
+                navigator.clipboard.writeText(text).then(mark_copied).catch(function() {
+                    fallback_copy(code, mark_copied);
+                });
+            } else {
+                fallback_copy(code, mark_copied);
```
senekor: Under what circumstances is this fallback needed? MDN docs say this API is widely supported: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard
I'd rather get rid of it if it's an unrealistic code path.

senekor (changes requested):
First pass, I'll have a closer look later.

This doesn't take translations into consideration, but those are on hold anyway, so there's not point in worrying about it now I guess.

I generally prefer styling be done via Tachyons instead of adding to the global app.scss file... but everything is nicely scoped to the `pre-copy-wrap` class, so might be fine.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/www.rust-lang.org/2281/3404e6e104f150ab08b95de85cfc81af4685d491..8ce508653b29b379d2ddc0a0279c45138b06b2d6)*

templates/components/tools/rustup.html.hbs
Mr-Sunglasses: Thanks for the review, I've added these as icon components.

static/scripts/tools-install.js
Mr-Sunglasses: You're right, clipboard API is widely supported, so no need of the fallback.

templates/components/tools/rustup.html.hbs
senekor: I mean an actual `.svg` file that goes in `static/images` and is included with and `img` tag.

static/scripts/tools-install.js
```diff
@@ -177,8 +177,38 @@ function check_initial_override() {
...
+            var code = wrap && wrap.querySelector('code');
+            if (!code) { return; }
+            var text = code.textContent;
+
+            var original_label = btn.getAttribute('aria-label');
+
+            function mark_copied() {
+                var text_el = btn.querySelector('.copy-code-btn__text');
+                if (text_el) { text_el.textContent = 'Copied!'; }
+                btn.setAttribute('aria-label', 'Copied!');
```
senekor, line 195: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/www.rust-lang.org/2281/3404e6e104f150ab08b95de85cfc81af4685d491..a7666fd4fa90d5f0f83806003785e0d2161fb094)*

I'm no expert in HTML accessibility... What is the point of setting the aria-label on an element that matches its text? I though aria-label was there to textually describe something that doesn't otherwise have a text representation, like an image.

Mr-Sunglasses: The button already uses `aria-label` as its accessible name (for example, “Copy install command to clipboard”), which gives more context than the visible text alone. Because `aria-label` can override the visible text for screen readers, when the button text temporarily changes to “Copied!” we also update the `aria-label` so assistive tech does not keep announcing the old label.


- Screen readers with setting the aria-label on an element that matches its text

https://github.com/user-attachments/assets/b8b02aaa-3957-4844-9105-452a57d10a32

- Screen readers without setting the aria-label on an element that matches its text

https://github.com/user-attachments/assets/7f25bf80-973e-480b-9c72-47b0ec07b414

templates/components/tools/rustup.html.hbs
Mr-Sunglasses: ah sorry my mistake I misunderstood, now i've changed them to `.svg` format and add via `img` tag

static/scripts/tools-install.js
senekor: makes sense!

senekor (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/www.rust-lang.org/2281/3404e6e104f150ab08b95de85cfc81af4685d491..d27267ed801b833f4b34c0234bb6824cbc0cec8f)*

PR #2282, opened 2026-04-11
Remove special text selection behavior of install command

senekor:
This reverts commit c7a310c50e5d6b6be9605d5fc6baa3791e0fc7d2.

Now that there is a copy button, there is no need for special selection behavior anymore.

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -115,10 +115,6 @@ code.code-header {
   margin: 0 0 3px 0;
 }
 
-code.copyable {
-  user-select: all;
-}
-
 .pre-copy-wrap {
   display: flex;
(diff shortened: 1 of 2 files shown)


Manishearth (approved):

Manishearth:
Actually, it's probably fine to keep both, some people will trust the visual more than a copy button

senekor:
One downside of the select all behavior is that it's not possible to copy part of the install command anymore. For example, someone might just want to read the install script or redirect it into a file before executing, instead of piping directly into a shell. This is a bit annoying at the moment. It's also pretty easy to select everything with a triple click, for manually copying.

Manishearth:
Ah good point!

PR #2283, opened 2026-04-12
Fix legacy URLs for working groups

### LINKED ISSUE #2268: Add redirects for working group pages

kaitersgonnakait:
An older version of the website had dedicated pages for working groups that do not exist anymore.

By filtering by teams that are subteam-of = "launching-pad" and kind = "working-group", the legacy pages for these working groups were determined to be:
* wg-cli at governance/wgs/cli
* wg-embedded at governance/wgs/embedded
* wg-gamedev at governance/wgs/gamedev
* wg-secure-code at governance/wgs/wg-secure-code

This PR creates redirects for these URLs to the new location of this information on the governance/teams page and also cleans up logic that was now unused for generating dedicated wg pages.

_Previously:_
~~When the site was converted from a Rocket server to a static generator, render_governance only iterated over top-level teams from index_data(), dropping pages for working groups under launching-pad (e.g. /governance/wgs/wg-secure-code). The page_data logic to preserve these URLs.~~

~~Adds launching_pad_subteam_pages() to produce the missing working group IndexTeam entries and chains them into the render loop.~~

~~By filtering by teams that are subteam-of = "launching-pad" and kind = "working-group", this creates the legacy pages for these working groups:~~
* ~~wg-cli at governance/wgs/cli~~
* ~~wg-embedded at governance/wgs/embedded~~
* ~~wg-gamedev at governance/wgs/gamedev~~
* ~~wg-secure-code at governance/wgs/wg-secure-code~~

~~An alternate solution to add redirects to the new location of the teams on this page here: https://rust-lang.org/governance/teams/~~

Closes #2268

### DIFF
--- a/src/redirect.rs
+++ b/src/redirect.rs
@@ -36,6 +36,16 @@ pub static PAGE_REDIRECTS: &[(&str, &str)] = &[
         "governance/wgs/wg-async",
         "governance/teams/lang#team-wg-async",
     ),
+    ("governance/wgs/cli", "governance/teams#team-wg-cli"),
+    (
+        "governance/wgs/embedded",
+        "governance/teams#team-wg-embedded",
+    ),
+    ("governance/wgs/gamedev", "governance/teams#team-wg-gamedev"),
+    (
+        "governance/wgs/wg-secure-code",
+        "governance/teams#team-wg-secure-code",
+    ),
     // miscellaneous
     ("governance/wgs", "governance#working-groups"),
     ("governance/all-team-members.html", "governance/people"),
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,6 +126,7 @@ So stay vigilant and update it if you notice something missing.
 - https://www.rust-lang.org/governance/teams/crates-io
 - https://www.rust-lang.org/governance/teams
 - https://www.rust-lang.org/governance/wgs
+- https://www.rust-lang.org/governance/wgs/wg-secure-code
 <!-- a couple localized redirects -->
 - https://www.rust-lang.org/es-ES
 - https://www.rust-lang.org/es-ES/contribute-compiler.html
--- a/src/render.rs
+++ b/src/render.rs
@@ -194,8 +194,13 @@ pub fn render_governance(
             .render(dst_path)
     })?;
 
-    // Individual teams
-    for team in data.teams {
+    // Individual teams, plus legacy launching-pad subteams that have dedicated pages
+    // for backwards compatibility (e.g. /governance/wgs/wg-secure-code).
+    let all_team_pages = data
+        .teams
+        .into_iter()
+        .chain(render_ctx.teams.launching_pad_subteam_pages());
+    for team in all_team_pages {
         println!("Rendering team page for {}", team.team.name);
         let data: PageData = render_ctx
             .teams
--- a/src/render.rs
+++ b/src/render.rs
@@ -194,13 +194,8 @@ pub fn render_governance(
             .render(dst_path)
     })?;
 
-    // Individual teams, plus legacy launching-pad subteams that have dedicated pages
-    // for backwards compatibility (e.g. /governance/wgs/wg-secure-code).
-    let all_team_pages = data
-        .teams
-        .into_iter()
-        .chain(render_ctx.teams.launching_pad_subteam_pages());
-    for team in all_team_pages {
+    // Individual teams
+    for team in data.teams {
         println!("Rendering team page for {}", team.team.name);
         let data: PageData = render_ctx
             .teams
--- a/src/teams.rs
+++ b/src/teams.rs
@@ -229,6 +229,28 @@ impl RustTeamData {
         })
     }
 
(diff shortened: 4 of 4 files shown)


senekor:
Ugh, I tried to reproduce where the regression happened. Turns out the old commits are completely broken, presumably because the team repo made breaking changes. 😢

I think I prefer redirects at this point. Or is there a specific reason we would prefer dedicated pages? I can't come up with one.

cc @Kobzol, do you have an opinion?

Kobzol:
I would definitely go for redirects, we don't have dedicated pages for all teams, and creating them only for a few WGs is inconsistent. I consider the fact that there were separate pages for a few special WGs to be a legacy implementation detail.

kaitersgonnakait:
Maybe I should have just closed this PR and opened a new one, but 🤷‍♀️

Anyway, made those legacy URLs redirects instead of dedicated pages!

src/redirect.rs
```diff
@@ -36,6 +36,16 @@ pub static PAGE_REDIRECTS: &[(&str, &str)] = &[
...
+    ("governance/wgs/cli", "governance/teams#team-wg-cli"),
+    (
+        "governance/wgs/embedded",
+        "governance/teams#team-wg-embedded",
+    ),
+    ("governance/wgs/gamedev", "governance/teams#team-wg-gamedev"),
+    (
+        "governance/wgs/wg-secure-code",
+        "governance/teams#team-wg-secure-code",
+    ),
```
senekor, line 48: Just so I can follow, why are there exactly these four redirects / subteams of launching-pad? The team repo has a bunch more launching-pad subteams, right?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/www.rust-lang.org/2283/09aa062a7c6467fd21d4d12e9949a5a69853b906..8ca096239a265b897ee3780a70436be0f8a23bab)*

senekor: Aah, it's only the working groups, right? Now I got it.

senekor: What about the archived wasm working group? Do we need a redirect for that as well?

kaitersgonnakait:
Here the timeline that I gathered from my code archaeology:

* 2018 - January 2025: Working Groups have a dedicated section on the /governance page, each working group has its own dedicated page. When a team is archived via PR in `team` repo, the dedicated page becomes unpublished.
* February 2025: The launching pad page is created, which contains sections for each working group. The Working groups section on the /governance page (which used to link to the dedicated working group pages) is removed. The PR indicates the old links to active working groups should continue to work for backwards compatibility reasons even though the launching pad page makes the dedicated wg pages obsolete, source: #2105 According to the [wayback machine](https://web.archive.org/web/20250302184009/https://www.rust-lang.org/governance), there were 6 working groups at this time: CLI, embedded, gamedev, wg-secure-code, wg-security-response, and wg-triage
* September 2025: Last known published version of dedicated working group pages according to [wayback machine] (https://web.archive.org/web/20250911054544/https://www.rust-lang.org/governance/wgs/wg-secure-code), I believe this PR (unintentially?) removed these legacy dedicated pages: #2174
* October 2025: An Archived Teams section is created on the website. The point of this change seems to be to republish info about teams that had previously become not visible on the site after being archived. No mention of preserving old URLs for backwards compatibility #2202 
* November 2025: wg-security-response and wg-triage was changed from a working group to a regular team. https://github.com/rust-lang/team/pull/2101/ and https://github.com/rust-lang/team/pull/2100/ Their old wg pages do not redirect anywhere.

My original goal with this PR was to correct the missing legacy dedicated pages that someone cared enough to maintain for a bit but seemed to have gone missing possibly by accident around September 2025, so that would mean just fixing those 4 working groups, but maybe there's a larger question here of how to handle when page location changes, either when there's a change in the www.rust-lang.org repo or in the teams repo.  It seems really cumbersome to maintain a whole hardcoded list of redirects for so many pages. Maybe having URLs that change sometimes is just the cost of having a website with dynamic data. It looks like the person who opened the issue was mostly concerned about the wg-secure-code page. Looks like that working group still links to the old URL in their GitHub profile, https://github.com/rust-secure-code, so maybe that URL should be updated.

Happy to update this PR to best reflect the philosophy of the core team.

src/redirect.rs
kaitersgonnakait: 🥴 Yes, those four are the currently active Launching Pad working groups. Good question about the wasm working group. It seems like the precedent for archived working groups or any archived subteams is to let the URL 404

senekor:
Thanks for that great investigation :slightly_smiling_face: Let's get these four redirects merged for now. The constantly changing data in the team repo does make it hard to keep all urls alive consistently. I can't think of a good solution to that atm.

senekor (approved):

PR #2299, opened 2026-05-21
Improve title of the person page

Kobzol:
I was googling "rust-lang <github username>" recently, and though Google managed to find the correct person page on the Rust website, the title kinda sucked (generic "Rust Project member"). This PR makes the title be the name of the person, which should look much better in search engine results.

### DIFF
--- a/locales/en-US/governance.ftl
+++ b/locales/en-US/governance.ftl
@@ -47,6 +47,6 @@ governance-all-teams-archived-title = Archived Rust Teams
 governance-all-teams-archived-intro = This section lists the { $count } archived teams that are no longer active. We want to thank all past members for their invaluable contributions!
 
 ## govenance/person.hbs
-governance-person-title = Rust Project team member
+governance-person-title = { $name }
 governance-person-team-member = Team member
 governance-person-team-alumni = Alumni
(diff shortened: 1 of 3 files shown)


locales/en-US/governance.ftl
```diff
@@ -47,6 +47,6 @@ governance-all-teams-archived-title = Archived Rust Teams
 governance-all-teams-archived-intro = This section lists the { $count } archived teams that are no longer active. We want to thank all past members for their invaluable contributions!
 
 ## govenance/person.hbs
-governance-person-title = Rust Project team member
+governance-person-title = { $name }
```
senekor, line 50: I don't get it, why is this going through the translation system? We're not gonna translate names, right?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/www.rust-lang.org/2299/9c75f61f37fde7d4e605232103b41cb003db462c..7465637e0fd05cdbef1240cd210c853582cb5ca2)*

Kobzol: It's not really about the translation, moreso about the fact that the title contains dynamic data. The code translates all titles and uses IDs to select the translation, so I added a mechanism to use ID + Fluent parameters. If I wanted to instead use a dynamic title, I'd have to add a mechanism to select between a translation ID or a normal string.

Up to you, what do you like more.

senekor: Ah, that makes sense.

senekor (approved):

PR #2323, opened 2026-07-25
Add dedicated Maintainer in Residence page

Kobzol:
This PR adds a new page to the website, which lists our Maintainers in Residence.

In theory, we could encode the MiR data in the `team` repo, to avoid putting the data part into the website. However, I think that it would be nice if we wrote some paragraphs of text about each MiR, similar to https://www.python.org/psf/developersinresidence/. That would likely be unwieldy to do in `team`, because it would have to live in a TOML file, and it would be separated from the website.

I went for a grid-like layout with two columns, to avoid having too wide descriptions of the invididual MiRs.

I filled in the first batch of MiRs:

image


Manishearth:
In favor, don't have time to review specifics

senekor:
Same, I should have time to review in August.

templates/funding/index.html.hbs
```diff
@@ -34,10 +34,16 @@
...
         {{/fluent}}
       </p>
 
-      <a href="https://github.com/sponsors/rustfoundation" class="button button-secondary mw6">{{fluent "funding-rfmf-donate"}}</a>
+      <div>
+        <a href="https://github.com/sponsors/rustfoundation" class="button button-secondary mw6">{{fluent "funding-rfmf-donate"}}</a>
+      </div>
+      <div class="pt3">
+        <a href="{{baseurl}}/funding/mir.html" class="button button-secondary mw6">{{fluent "funding-rfmf-mir-link"}}</a>
+      </div>
```
senekor, line 48: Would putting these side by side look nicer?

Image

src/mir.rs
```diff
@@ -0,0 +1,91 @@
...
+
+        let login_differs_from_name = person.name != login;
+        Self {
+            login,
+            name: person.name.clone(),
+            teams: teams.join(", "),
+            role: match role {
+                MirRole::FullTime => "Full-time Maintainer in Residence",
+                MirRole::PartTime => "Part-time Maintainer in Residence",
+                MirRole::Grant => "Maintainer grant",
```
senekor, line 49: non-blocking: These should probably be translation keys? But translations are on hold anyway.

src/mir.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// and Maintainer grants.
+    pub fn build(team_data: &'a RustTeamData) -> Self {
+        let mirs = vec![FundedMaintainer::new(
+            team_data,
+            "LoriLorusso",
+            &["all"],
+            MirRole::FullTime,
+            r#"
+Lori Lorusso works as a full-time Rust Foundation employee.
+"#,
```
senekor, line 86: note: I'm okay with keeping this data in the repo. Are we expecting these notes to be translated? In that case, it would live in the translation files. Otherwise we can technically keep it somewhere else, but I can't think of something better anyway.

senekor (commented):
All good so far 👍 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/www.rust-lang.org/2323/7e9b72fe18179a12562f75a3ee523516ad02ce89..3f59376ec80a54a28690eaf2f8e38ccca608e483)*

src/mir.rs
Kobzol: I tried to do it, but I actually found it really tricky to produce translations from Rust code. The handlebars_fluent crate does not accept dynamic translation keys, the key has to be a string constant.

src/mir.rs
Kobzol: I don't expect them to be translated, to be honest, it would be a lot of work. (and see my other comment on why translation from Rust is an issue)

templates/funding/index.html.hbs
Kobzol: image

Possibly, yeah. WDYT?

senekor: Yeah, I like this. Maybe with a slightly bigger gap. Looks a little crammed right now.

src/mir.rs
senekor: Right, let's keep it like this then. No need to prepare for translations that might never materialize.

src/mir.rs
senekor: Oh, are you thinking to just keep the real data right in this rust file? Actually, fine by me as well.

templates/funding/index.html.hbs
Kobzol: image

Kobzol:
Thanks for the review! I will keep this drafted until we actually fill it with the first batch of MiRs, because it doesn't really make sense to have an empty page there.

src/mir.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// and Maintainer grants.
+    pub fn build(team_data: &'a RustTeamData) -> Self {
+        // MiRs
+        let rami3l = FundedMaintainer::new(
+            team_data,
+            "rami3l",
+            &["rustup"],
+            MirRole::FullTime,
+            r#"
+rami3l has been a member of the Rustup team since 2023, and its lead since 2025. He deeply cares about developer tooling, and the facets of Rust that many might take for granted. In addition to maintaining the Rustup tool, he also introduces new contributors to the Rustup team, and has mentored two Google Summer of Code projects, in 2024 and 2025.
```
rami3l: One tiny typo :)

```suggestion
rami3l has been a member of the Rustup team since 2023, and its lead since 2025. He deeply cares about developer tooling, and the facets of Rust that many might take for granted. In addition to maintaining the Rustup tool, he also introduces new contributors to the Rustup team, and has mentored two Google Summer of Code projects, in 2025 and 2026.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/www.rust-lang.org/2323/2ca00346fbe04bbb072c2ef26c2bea15b9f087f1..d36fbb79120c38da376fd0710eb31ca0d5d063dc)*

senekor:
Is there a specific reason to use different photos stored in this repo? Why not use the GitHub profile photos? They're not very big, so it's not a huge issue, but I guess we're expecting these maintainers to change on a somewhat regular basis. Adding and deleting photos all the time will bloat the repo.

Kobzol:
Many people have a sort of opaque or auto generated GitHub profile picture, so in this page I wanted it be a bit more personal and let people choose custom photos. I tried to reduce their size, right now, it is ~380 KiB combined, and if I remove rami3l, Chris and Jarcho's pictures, who just have their GitHub profile picture there anyway, the size would be ~280 KiB.

The images directory without the `mir` subdirectory takes 3.2 MiB currently, so adding an additional ~300 KiB doesn't sound that bad. Even if we had 5 new MiRs every year, which is most likely extremely optimistic, adding 300 KiB per year to the repository doesn't sound like such a large bloat.

> but I guess we're expecting these maintainers to change on a somewhat regular basis.

Not necessarily, the happy path is that new funded maintainers will be added, but it's not like we're gonna replace everyone every year, it is supposed to be long-term stable funding.

That being said, if you don't want to have the pictures here, I can just replace them with the GitHub avatar links.

senekor:
Makes sense, I'm ok with keeping the more personal pictures in-repo then. Can we use the GitHub profile pictures for the people where it's the same anyway? Or would that be too complicated implementation-wise?

Kobzol:
It's actually already implemented, I just needed to remove the images. Done that now.

senekor (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/www.rust-lang.org/2323/2ca00346fbe04bbb072c2ef26c2bea15b9f087f1..88252836474ec9fe50781b7dd26d87e4785b38c5)*

Kobzol:
I'd merge this next week, after we give a bit more time for MiRs to modify the descriptions, and after we merge the [blog post](https://github.com/rust-lang/blog.rust-lang.org/pull/1914).

joboet (approved):

Kobzol:
I squashed the photos to remove unused photos from git history. The blog post will be merged today, so you can merge this whenever you want. Thank you very much!

PR #2332, opened 2026-09-01
goodbye RFC reference

lcnr:
see https://rust-lang.zulipchat.com/#narrow/channel/392734-council

RFCs should not be the entry-point for new contributors and users to Rust. Teams have their own lower overhead processes and suggesting users to start engaging with the Project by writing fully-fleshed out RFCs is actively harmful imo

unsure how to improve this section given the status quo and better documenting that is likely part of https://rust-lang.org/governance/teams/launching-pad/#team-comprehensibility

### DIFF
--- a/locales/en-US/governance.ftl
+++ b/locales/en-US/governance.ftl
@@ -3,10 +3,6 @@
 ## governance/index.hbs
 governance-page-title = Governance
 governance-tagline = How Rust is built by its community
-governance-rfc = RFC process
-governance-rfc-blurb = Each major decision in Rust starts as a Request for Comments (RFC).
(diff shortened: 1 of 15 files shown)


clarfonthey:
Considering how there are translations for this blurb, it would make sense at least to redo it and replace it, so, I think removing it for now is fine.

Manishearth:
As stated in Zulip, I would prefer we replace this with a more accurate short picture of the Rust govenrnace model, with the goal of communicating:

 - we are open governance
 - you can impact governance
 - here is how to get involved
 - here are some of our processes

The section can have multiple link buttons at the bottom.

